PR #42, opened 2015-09-19
Colonelfazackerley rebase

tttppp:
This should be what colonelfazackerley has, but with the two correction commits smoothed out (and the subsequent merge).

I've not tried running anything, and I'd be happy for @colonelfazackerley to fetch my changes and re-push them as a merge request of his own.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -2,7 +2,7 @@ Time moves upwards.
 
 Later
 -----
-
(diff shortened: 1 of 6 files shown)


tttppp:
No longer needed as colonelfazackerley has resolved the problems.

PR #48, opened 2015-09-23
initial doxygen config

colonelfazackerley:
result can be seen here

http://www.extant.plus.com/rabbit_doxy/index.html

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,0 +1,1459 @@
+# Doxyfile 1.8.8
+
+# ColonelFazackerley says install doxygen and graphviz. then "doxygen Doxyfile"
+
+# This file describes the settings to be used by the documentation system
+# doxygen (www.doxygen.org) for a project.
+#
+# All text after a double hash (##) is considered a comment and is placed in
+# front of the TAG it is preceding.
+#
+# All text after a single hash (#) is considered a comment and will be ignored.
+# The format is:
+# TAG = value [value, ...]
+# For lists, items can also be appended using:
+# TAG += value [value, ...]
+# Values that contain spaces should be placed between quotes (\" \").
+
+#---------------------------------------------------------------------------
+# Project related configuration options
+#---------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


andybalaam:
Not convinced the graphs add much.  What do you think?

andybalaam:
Shall I host the generated docs somewhere on artificialworlds?

colonelfazackerley:
I like the graphs. Probably becuase I am struggling to remember how Java OOP works. I think it's possible to tweak the config so it does not generate so many.

On your site would be best. I would happy to help with keeping that up to date, if I can.

andybalaam:
Cool, if they're helpful, they can stay.  I'll make a Makefile target to push them to my site and do it when I remember.  If we need to automate it some day, we can.

andybalaam:
It's here http://artificialworlds.net/rabbit-escape/doxygen/ - let me know if it's not ok.

colonelfazackerley:
I wrote this to push to my ISP. My ISP only has a rather naff service, ftp upload only.

```
#!/bin/bash    
HOST="ftp.plus.net"
USER="extant"
echo -n "Password for $USER:" 
read -s PASSWORD
FTPURL="ftp://$USER:$PASSWORD@$HOST"
LCD="./doc/html/"
RCD="/htdocs/rabbit_doxy"
DELETE="--delete"
lftp -c "set ftp:list-options -a;
open '$FTPURL';
lcd $LCD;
cd $RCD;
mirror --reverse \
       $DELETE \
       --verbose "

```
I excluded it from the pull request, as having so much config specific to me didn't seem right.

andybalaam:
See the target doxygen-upload I put in the Makefile for my equivalent.  Hopefully not too noisy although it clearly can only be run by me.

colonelfazackerley:
The link you put in INSTALL.md works.

rsync is much nicer than what I did. I could not use it for this, as SSH is needed on the server end, I think.

andybalaam:
Yes, rsync uses ssh by default, which does require a server.

PR #54, opened 2015-09-25
No duplicate properties in world metadata

colonelfazackerley:
The lack of this caused me some wasted time, when I had rabbit_delay set twice in a file and did not notice. It just used the last value.

I found a duplicate block in easy 15.

I tested, got the whitespace right, merged upstream commits in first. Think I am getting the hang of this... (probably means I am getting more dangerous)


tttppp:
Can we still have duplicate lines starting
```
:*=
```
This is used when multiple items are on the same tile.

colonelfazackerley:
They are called starPoints in the code, I think. They are not affected.
It's just stuff like
```
...
:name=a
...
:name=b
...
```
and
```
...
:dig=1
...
:dig=3
...
```
which will be caught.

PR #65, opened 2015-09-28
Author key

colonelfazackerley:
code to display the author for the level. also added the author key to easy, medium and hard sets.


andybalaam:
Looks really good, thanks.  Please can you rebase?  I still have no proper Internet, but may be able to apply this via my phone.

SwingGameLaunch.java line 196 - missing space before the equals sign.

What do you think should be the standard author format?  A name + (optional?) url?  How should we display them?  Do we need authorName and authorUrl fields?

andybalaam:
Were you planning to use GitHub ids?  That's quite nice, but no good for people who don't have a GitHub account.  I think we want a plan text name and a URL, which could be to a GitHub profile, or somewhere else.

If you agree, I think the best way to do that would be 2 fields: authorName and authorUrl.  What do you think?

tttppp:
I'd be happy with any of authorName only, nothing at all, a shout out in the "about" section. If you want to use a name then TTTPPP sounds good.  If you add URLs then please can they be made hard to accidentally click - as I wouldn't want to visit a random github profile while I'm trying to get my rabbit-saving fix!

colonelfazackerley:
I agree that an optional URL, possibly to a github page would be best. I can update the pull request with that. 

I did rebase... What should be the aim of further rebasing? I did accidentally include one of your commits in the rebase range. Not sure I know how to undo that.

andybalaam:
Every time I commit something conflicting onto master, you need to rebase to bring this up to date with the latest.  I'm also not sure why one of my commits is now listed here.  I will try a rebase here.

Just for reference, here is what I would do if I were you:

    git checkout master
    git pull upstream master     # If that fails because you have committed onto master, then git pull --rebase upstream master
    git checkout author-key
    git rebase master
    git push -f origin author-key

andybalaam:
I have rebased and merged this manually, thank you!

PR #122, opened 2015-10-25
Mega coder

tttppp:
Obfuscation of hints and solutions.  Uses a suffix ".code" to allow any string in a level file to be obfuscated.  Currently creating the encoded strings is a pain, but there's a unit test file that can do it.

Also worth noting is the license on the MegaCoder. I didn't feel comfortable restricting this to just GPL use - what if a closed source company was in a pinch and needed string obfuscation?  I wouldn't want them to have to deliver their strings in the clear!

Also, do you have a code formatter for Eclipse? Doing formatting by hand is a pain!

### DIFF
--- a/rabbit-escape-engine/test/rabbitescape/engine/TestTextWorldManip.java
+++ b/rabbit-escape-engine/test/rabbitescape/engine/TestTextWorldManip.java
@@ -2,6 +2,7 @@ package rabbitescape.engine;
 
 import static org.hamcrest.MatcherAssert.*;
 import static org.hamcrest.core.IsEqual.*;
+import static org.junit.Assert.assertEquals;
 import static rabbitescape.engine.ChangeDescription.State.*;
 import static rabbitescape.engine.Tools.*;
 import static rabbitescape.engine.textworld.TextWorldManip.*;
@@ -457,6 +458,21 @@ public class TestTextWorldManip
         );
     }
 
+    @Test
+    public void Can_obfuscate_hints()
+    {
+        String[] lines = {
+            ":hint1.code=_Uf?>3ZH_8>U>{3",
+            ":hint2.code=@XW@W:)e+:+ ",
+            ":hint3.code=X5g..T[6X4["
+        };
+
+        World world = createWorld( lines );
+
+        assertEquals( world.hint1, "Select the bash" );
+        assertEquals( world.hint2, "Use the bash" );
+        assertEquals( world.hint3, "Be the bash" );
+    }
 
     @Test
     public void Can_provide_number_of_rabbits()
--- a/rabbit-escape-engine/src/rabbitescape/engine/textworld/LineProcessor.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/textworld/LineProcessor.java
@@ -17,10 +17,13 @@ import rabbitescape.engine.Rabbit;
 import rabbitescape.engine.Thing;
 import rabbitescape.engine.Token;
 import rabbitescape.engine.util.Dimension;
+import rabbitescape.engine.util.MegaCoder;
 import rabbitescape.engine.util.VariantGenerator;
 
 class LineProcessor
 {
+    private static final String CODE_SUFFIX = ".code";
+
     private static class Point
     {
         public final int x;
@@ -175,6 +178,12 @@ class LineProcessor
 
         String key   = splitLine[0];
         String value = splitLine[1];
+
+        if ( key.endsWith( CODE_SUFFIX ) )
+        {
+            key = key.substring( 0, key.length() - CODE_SUFFIX.length() );
+            value = MegaCoder.decode( value );
+        }
 
         if ( TextWorldManip.META_INTS.contains( key ) )
         {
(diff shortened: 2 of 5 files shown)


colonelfazackerley:
I look forward to looking at the MegaCoder.

eclipse format
https://github.com/andybalaam/rabbit-escape/blob/master/doc/eclipse-format.xml

tttppp:
Excellent, many thanks! That will save me a world of pain adding spaces around brackets next time :-)

rabbit-escape-engine/test/rabbitescape/engine/TestTextWorldManip.java
```diff
@@ -457,6 +458,21 @@ public void Can_provide_empty_description()
...
+    {
+        String[] lines = {
+            ":hint1.code=_Uf?>3ZH_8>U>{3",
+            ":hint2.code=@XW@W:)e+:+ ",
+            ":hint3.code=X5g..T[6X4["
+        };
+
+        World world = createWorld( lines );
+
+        assertEquals( world.hint1, "Select the bash" );
```
colonelfazackerley: I would use the hamcrest asserts instead. They are nicer, in that when they fail, they shoe you the 2 non matching strings.
```java
assertThat( world.hint1, equalTo("Select the bash") );
```

tttppp: @colonelfazackerley junit.Assert.assertEquals also gives useful messages on failures. The main reason for using it was that I have junit.Assert set up with autosuggest in Eclipse.

andybalaam:
By convention in rabbit escape we are using hamcrest asserts so I'd prefer that unless you have a good reason to do otherwise.

andybalaam:
Why `key.code=blah` rather than `key=code:blah` ?

andybalaam:
Advantages: a value can start with anything, without having to avoid the magic string. 

Disadvantages: it kind of looks like you are allowed to supply coded and uncoded properties. I haven't looked at the code yet so I haven't seen how you handle this. 

Opinions?

andybalaam:
I guess I asked for this by being so anti-dependencies but any excuses for making up our own encoding rather than base64? I acknowledge that it is fun...

andybalaam:
Please can you document the obfuscation algorithm?

andybalaam:
Please rebase into a clean commit.

andybalaam:
Also, you may enjoy considering whether the entire algorithm may be implemented as in-place swaps on a single array of char, avoiding any memory allocations.

tttppp:
@andybalaam Most of your comments are just in the general comments thread, making them confusing.

> Why key.code=blah rather than key=code:blah?
As you noted starting value with "code:" removes the possibility of a value actually starting with "code:". The standard validation kicks in if you include two copies of a single-value string (e.g. :description and :description.code).

> Why not base64?
Surely no one wants to try to write a base64 encoder for fun!  Another advantage of the MegaCoder is that the range and domain are equal, so strings are no longer than they need to be.

> Documentation of algorithm
There's a brief summary in the docstring at the top of the MegaCoder, but I can try to nail it down a bit.

> Please rebase into a clean commit
I don't know what you mean by "clean commit".  Do you just mean make changes in-place in the history?

tttppp:
Some of the review comments have been addressed.  Some more can be done when I have another chance to look at this.

Maybe I'm using github incorrectly, but this thread seems really hard to follow!

andybalaam:
@tttppp several of my comments were targeted at individual lines of code. I agree that pull requests can be hard to follow. Maybe we will learn how to do it better over time.

andybalaam:
Having read the code I am happy with the .code key suffix.

andybalaam:
I feel sure plenty of fun could be had writing a base64 encoder but one must follow one's own sense of fun.

andybalaam:
By clean commit I meant a single commit containing the change including all fixes.

andybalaam:
Yes, shorter is better so that is a good enough excuse not to use base64.

andybalaam:
Sorry, I misspoke slightly about clean commits. I like having 2 separate commits for the MegaCoder and its use in level parsing,  but what I would like is for the corrections and changes due to review comments to be merged into the original MegaCoder commit rather than being extra commits later in history. If this is a problem we can chat about it further.

tttppp:
A couple of commits sounds reasonable for pretending I'd got everything right first time ;-)  I don't think there's a way to rebase a pull request without creating a new one, so I suggest tidying up the history is the last thing we do with this.  Probably with a larger pull request like https://github.com/andybalaam/rabbit-escape/pull/123 it's not worth the effort of trying to tidy up history.  I think the pull request model is built with incremental changes in mind, rather than squashing new changes into the existing commits.

On a larger project there would be the fun issue of people having merged commits from an open feature branch into their own branch, and changing history would be pretty awful then.  With a small number of devs I think squashing existing changes could work fine (I often do this locally).

andybalaam:
I do prefer squashed commits that look like you did it right first time because it's easier to read back through history. To do this on a pull request you just `git push -f branch_name`

andybalaam:
Sorry,  `git push -f origin branch_name`

andybalaam:
(Having already rebased your local branch)

andybalaam:
Brilliant, thanks.

PR #123, opened 2015-10-26
GitHub issue level get

colonelfazackerley:
There is still plenty to do . I don't mind if you fix stuff, or ask me to make more pull requests.

I would like this to be merged.


andybalaam:
Merged since this is all good stuff and we don't want to get too diverged.  Let's address any other review comments etc. in new commits/merges.

PR #124, opened 2015-10-27
bumping head when on slope turns rabbit. fixes #68 and fixes #91.

### LINKED ISSUE #68: Bug?: Rabbit walks up bridge, should bump head
```
###   ###  
  (#     # 
r(# O ri# O
##### #####
```
-Rabbit on the left walks up bridge, but should bump head, and turn around.
-Rabbit on the right correctly can't build the bridge.

The only way the bridge could get like that, is in a level design. So, it might be considered a bug in the level, instead of a bug in the rabbit movement.

The parser could just check for solid blocks above bridges, and forbid them. However, sometimes they just look nice (#67 example). So, I think it would be better to check the rabbit's movement and make them turn around.

### LINKED ISSUE #91: Bug: Rabbit ignores Block above him
Hey,

I detected a possible bug in level 7 (medium):
```
######################
#                /#  #
#                / (O#
#                /# ##
#               (/#  #
#              (/##/)#
#             ( /#\# #
#            (    # /#
# Q         (    #/###
#          (     ##\##
#     r   (     /###/#
######################
```

After climbing a Triangle-Block ('/') the rabbit ignores the block above him and goes further instead of turning around.

lvl7

The block above is also ignored when it's a full block.

colonelfazackerley:


andybalaam:
Lovely simple fix. Really rubs in the need for us to be able to test that all levels are soluble after behaviour changes (which I am working on).

I'd prefer to assert that the rabbits have turned around rather than they haven't not turned around. E.g. If you made a typo in the badLines string anywhere the test would pass even if the bug were not fixed. 

I'd mildly prefer the test to be in TestWalking but could be persuaded otherwise.

colonelfazackerley:
i updated the branch. 
- Another test in the other sense?
- I was thinking it might be nice to have a list of bug fix tests, but I am not all bothered. I moved it.

andybalaam:
Great, thanks.

PR #125, opened 2015-10-28
falling rabbits not allowed to change to exit state. fixes #8

### LINKED ISSUE #8: Rabbits enter exit when they should die from falling
If a rabbit falls too far, but lands on the exit, it escapes.  Actually, it should die.

colonelfazackerley:


andybalaam:
Should the change be in the code that decides the exiting behaviour has been triggered? On my phone so can't check properly.

colonelfazackerley:
The test should be in checkTriggered instead? I am happy to wait until you can look at it properly. I don't have a good grasp of the behaviour code. Some doxygen on it would help me.

andybalaam:
Yes, the change should be in checkTriggered. checkTriggered decides whether this behaviour is active and newState decides what to actually do. I think the activeness (triggeredness) affects other actions' decisions about whether to become active themselves.

colonelfazackerley:
changed and rebased.

andybalaam:
Nice fix, ta.

Issue #102, opened 2015-10-13
Bug: keystrokes lost


colonelfazackerley:
After clicking start to clear the dialog and start the game, keystrokes are lost. Probably because the keyboard focus is not in the right place.

andybalaam:
Needs more research.  It may not be possible to fix.

PR #129, opened 2015-10-30
grab keyboard focus after dialog goes so keystrokes are not lost. fix…

colonelfazackerley:
…es #102 

Seems clear that after the dialog is done the focus should return to the parent. This code should not really be necessary.


andybalaam:
These small bug fixes are incredibly valuable, thanks!

PR #136, opened 2015-11-01
Fall and walk past exit

colonelfazackerley:
I forgot to add the fix # in the commit message.


andybalaam:
Yes, in retrospect this is much better.
I think there is already a method somewhere in test code to step a world several times. I'd prefer to use that rather than add a method to World that is only used in tests.

colonelfazackerley:
i will try and find that other method this evening.

colonelfazackerley:
there is TestEntrance.fiveSteps()

but I think mine is better...

colonelfazackerley:
WorldAssertions.assertWorldEvolvesLike()

looks better. I will look into it.

andybalaam:
Yes I like that. It forces you to show all steps, which I basically like, but could probably be persuaded to allow skipping steps.

PR #152, opened 2015-11-04
fixes #100 rabbits do not consume tokens while splatting.

### LINKED ISSUE #100: Bug: splat from height consumes token
```
r  
#  
   
   
   
 i 
###
```
This rabbit should not consume the bridge token as they die.

colonelfazackerley:
fixes #100.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
@@ -31,8 +31,31 @@ public class BehaviourTools
         return pickUpToken( type, false );
     }
 
+    public boolean rabbitIsFalling()
+    {
+        return
+               State.RABBIT_FALLING == rabbit.state
+            || State.RABBIT_FALLING_1 == rabbit.state
+            || State.RABBIT_FALLING_1_TO_DEATH == rabbit.state
+            || State.RABBIT_DYING_OF_FALLING_2 == rabbit.state
+            || State.RABBIT_DYING_OF_FALLING == rabbit.state
+            || State.RABBIT_FALLING_ONTO_LOWER_RIGHT == rabbit.state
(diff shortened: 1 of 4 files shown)


andybalaam:
I think you can ask the falling behaviour whether it's been triggered to find out whether the rabbit is falling. I need to think about this a bit more - it seems weird. Couldn't we just ask whether it's falling to death based on its state and that should be enough?

colonelfazackerley:
"Couldn't we just ask whether it's falling to death based on its state and that should be enough?"
No, that does not work. It's the first thing I tried. I think the reason it does not work, is to do with falling 2 squares per time step.

 "I think you can ask the falling behaviour whether it's been triggered to find out whether the rabbit is falling"
I will look at that. It's still tricky to get at the rabbit's Falling instance inside the behaviours List.

colonelfazackerley:
see #155 about using state to check for falling to death.

I have also just tried using Falling.checkTriggered() to test for falling instead of that method that lists all the falling states. Falling.checkTriggered() returns false if there is a block below. When the rabbits are on the same square as the token, there is a block below, so checkTriggered says they are not falling and they gobble tokens.

There may be a neater way of doing this, but those don't work.

andybalaam:
Hmm. See my comments on #155. If I get a chance soon to think properly about this and try some things out I will get back to you.If not I will merge this and come back to it. I think baddy rabbits will force us to find a cleaner system.

andybalaam:
Merging since the behaviour is clearly correct and we can refactor the implementation later if we find a better way.

colonelfazackerley:
I think it is valuable having a set of tests that cover situations that are bound to be awkward, however rabbit movement is coded.

I don't mind at all if the tests i have contributed are the only thing that survives a rewrite.

andybalaam:
Couldn't agree more.

andybalaam:
(As a general principle, not meaning that I want to throw away this code :-)

PR #155, opened 2015-11-05
Dont merge this!

colonelfazackerley:
I just did a pull request so you can see easily what happens if this is done wrong.

In the test case the rabbits are falling different heights. A rabbit falling an even number of squares splats and consumes a token.
```
    [junit] Testcase: Rabbits_falling_to_death_do_not_consume_tokens took 0.022 sec
    [junit] 	FAILED
    [junit] 
    [junit] Expected:     ["         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "#bdi    #", "####****#", "#########", ":*=)k", ":*=(c", ":*=\p", ":*=/db"]
    [junit]      but: was ["         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "#b i    #", "####*(**#", "#########", ":*=)k", ":*=\p", ":*=/b"]
    [junit] junit.framework.AssertionFailedError: 
    [junit] Expected:     ["         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "#bdi    #", "####****#", "#########", ":*=)k", ":*=(c", ":*=\p", ":*=/db"]
    [junit]      but: was ["         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "         ", "#b i    #", "####*(**#", "#########", ":*=)k", ":*=\p", ":*=/b"]
    [junit] 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
    [junit] 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
    [junit] 	at rabbitescape.engine.util.WorldAssertions.assertWorldEvolvesLike(WorldAssertions.java:148)
    [junit] 	at rabbitescape.engine.logic.TestTokens.Rabbits_falling_to_death_do_not_consume_tokens(TestTokens.java:288)

```

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
@@ -31,27 +31,17 @@ public class BehaviourTools
         return pickUpToken( type, false );
     }
 
-    public boolean rabbitIsFalling()
+    public boolean rabbitIsFallingToDeath()
     {
         return
-               State.RABBIT_FALLING == rabbit.state
-            || State.RABBIT_FALLING_1 == rabbit.state
-            || State.RABBIT_FALLING_1_TO_DEATH == rabbit.state
+               State.RABBIT_FALLING_1_TO_DEATH == rabbit.state
             || State.RABBIT_DYING_OF_FALLING_2 == rabbit.state
-            || State.RABBIT_DYING_OF_FALLING == rabbit.state
-            || State.RABBIT_FALLING_ONTO_LOWER_RIGHT == rabbit.state
-            || State.RABBIT_FALLING_ONTO_RISE_RIGHT == rabbit.state
-            || State.RABBIT_FALLING_ONTO_LOWER_LEFT == rabbit.state
-            || State.RABBIT_FALLING_ONTO_RISE_LEFT == rabbit.state
-            || State.RABBIT_FALLING_1_ONTO_LOWER_RIGHT == rabbit.state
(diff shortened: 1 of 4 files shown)


andybalaam:
Interesting. I am starting to think we need to break up the behaviour into more steps. Maybe:
1. Decide which behaviour is triggered for each rabbit
2. Decide the next state of each rabbit
3. Decide the next location of each rabbit
4. Modify next states and locations based on interactions e.g. collide with baddy rabbit
5. Set states to next state

Then we animate the states.

6. Move rabbits to next position

And repeat.

I'm not sure where picking up tokens fits in. Thoughts welcome.

colonelfazackerley:
I am going to call your numbered items substeps for clarity, we already have steps.

I guess the rabbit could decide if it is going to pick up a token in the previous step's no.3 substep. Would have to then see which rabbit wins if multiple rabbits arrive at a token location in the same step.

Not sure this will work out any simpler than what we have now.

Can you merge #152 in the interim? The tests are decent, and it works, even if it could be neater.

colonelfazackerley:
Aims for refactoring:
- Handle rabbits on slopes cleanly. It's different to rabbits on flat topped blocks in that they are in the same location.
- Handle falling cleanly. Moving two squares per step is the cause of complexity in the code and some bugs.
- Reduce the number of rabbit states. Maybe use booleans to hold some state.

andybalaam:
Agree on all points.

Issue #144, opened 2015-11-02
Enhancement: Solution simplifier


tttppp:
When recording a solution using the command line game interface, the solutions are very verbose. It would be good to combine wait instructions in particular, but also strip out extraneous changes of selected token.  There might be other things too which could be simplified?

I've made a start on this in https://github.com/tttppp/rabbit-escape/tree/recordingSimplifier and if anyone wants to get it  to a state where it can be merged then just add a note in this thread.  I think it needs the output string constructing and a test.

andybalaam:
I haven't looked at this yet, but as a general comment I'd prefer to be constructing a list of `Instruction`s and converting that back into a string.  It would make this cleaner, and be better style.

tttppp:
I felt similarly, so that's what I've done. The downside is that it requires converting back to a string at the end, which is one of the outstanding issues I mentioned above.

I'm hoping we can get this in before too many solutions are created, as it should make much shorter solutions that are less memory intensive. I guess we could always apply it retrospectively.

colonelfazackerley:
I see what you mean

:solution.1=8;bridge;;;(12,7);;(12,7);;;;;;;;(10,6);;;;;;;(13,4);;;;;4;2;2;2;2;;;1
:solution.1.code="Rccd6cc!c!c!d8(cRhccccd?dch!cc(!c8ZcRc(clccgOc^cRu(cC6c~cdc!ccchccchc

for now, i intend to manually clean up the solution. if i do this for a bunch of levels. what's the best way to obfuscate? (We will have the same issue with the hints).

andybalaam:
The production levels should not be obfuscated. We only need obfuscation when sharing levels with each other in source form.

andybalaam:
Is this done?

andybalaam:
Yes, it is :-)

PR #157, opened 2015-11-06
Make solutions a mini-interpreter

### LINKED ISSUE #148: Bug: after winning text mode, the reported solution does not work
This is a reported solution for hard 1
```
:solution.1=8;bridge;1;1;(12,7);1;(12,7);1;1;1;1;1;1;1;(10,6);1;1;1;1;1;1;(13,4);1;1;1;1;4;2;2;2;2;1;1;1;1;1;1;1;1;
```
(plus/minus a bit of waiting). I put some println in world.step() to see what happened.
```
    [junit] RUNNING saved0 killed0
    [junit] RUNNING saved0 killed0
   <snip>
    [junit] RUNNING saved0 killed0
    [junit] RUNNING saved0 killed0
    [junit] RUNNING saved0 killed1
    [junit] RUNNING saved0 killed1
    [junit] RUNNING saved0 killed2
    [junit] RUNNING saved0 killed2
    [junit] RUNNING saved0 killed3
    [junit] RUNNING saved0 killed3
```
They are dying in there!  but I can't see how.

andybalaam:
@colonelfazackerley @tttppp I am getting uncomfortable with not having proper review of my code.

I will be pushing to this request over this evening, and would appreciate your comments now or later.  I'll leave time for at least one person to review before merging (with fixes where needed).

The purpose of this pull request is to convert solutions into a little programming language:

- String solution gets parsed into a Solution, which contains Instructions that map 1-1 with the String language.  Solutions and Instructions know nothing about Worlds.

- The Solution is then interpreted by the SolutionInterpreter, which is an Iterable<SolutionTimeStep>, allowing us to move through the solution step by step.  Every time you iterate the SolutionInterpreter, you are expected to call World.step() too, but we don't do it for you, allowing things like the SwingUI to use the SolutionInterpreter without becoming subservient to it.

- The SolutionTimeSteps know how to modify a world based on the Instructions in the solution.

- SolutionInterpreter will need to be iterated 10 times to represent a single WaitInstruction( 10 ).

- This will hopefully clean up the code, fix #148, and allow us easily to make Solutions round-trip cleanly back into strings.


colonelfazackerley:
I have been trying to get a copy of this to try.
```
git co --track  -b upstream/iterable-solutions
```
but then i 
```
git log -3
```
and all I can see is commits from a couple of days ago.

tttppp:
@colonelfazackerley I think the answer is don't use the -b:
```
git checkout -t upstream/iterable-solutions
```

tttppp:
@andybalaam I didn't get very far with my branch. I haven't had much chance to revisit it:
https://github.com/tttppp/rabbit-escape/commit/bc0b8f51dd28685a7ec61b9fdd2133fa9534a3d3

I originally mapped solution strings to lists of lists of instructions (i.e. lists of steps containing lists of instructions: dig&(1,3);bash&(6,2)...)  I have since thought this is a mistake, and we should just have lists of instructions, because there is an instruction to wait.

I hope you can get something working, and if you can add in the simplification too (#144) all the better!

colonelfazackerley:
I managed to pull this branch
```
cf@re:~/re/rabbit-escape$ git co -b iterable-solutions
Switched to a new branch 'iterable-solutions'
cf@re:~/re/rabbit-escape$ git pull upstream iterable-solutions
From https://github.com/andybalaam/rabbit-escape
 * branch            iterable-solutions -> FETCH_HEAD
Updating e71d383..0d6fae7
Fast-forward
<snip>
```
but now I need to sleep. I will have a look tomorrow.

andybalaam:
@colonelfazackerley sorry, I make this pull request as I started work, so there wasn't much to see, but maybe by the morning there will be more.

andybalaam:
@tttppp good to know I won't be treading on your toes.  I too vacilated about each action being a list of actions, but I think the idea of working in time steps makes it all make sense.  If/when I get it working I hope you'll agree :-)

andybalaam:
@colonelfazackerley if you just want to look at the code I suggest the "commits" tab above.  If/when I actually do something useful instead of refactoring, obviously you will need to pull the code to try it.

andybalaam:
Also please note I may "force push", changing history, if I screw something up, or there are fixes coming out of the review.  If I do this (I have already done it once) you will need to trash your branch and re-pull.  I'm not certain how to do this if you're just looking at it as a remote.

andybalaam:
Ah, I see you are pulling it as a proper branch.  In that case, if I screw you you should be able to:

    git pull -f upstream iterable-solutions

I think, or just delete and recreate:

    git branch -d iterable-solutions
    git co -b iterable-solutions
    git pull upstream iterable-solutions

andybalaam:
With my latest commit the interpreter basically works.  When I tried it, it fixed the problem with easy 1 where we had to append an extra ";1".

TODO:
- annotate each produced timestep saying what "solution step" it came from (to improve errors)
- rename "solution step" to something better
- hook up to where we run solutions and iron out any problems
- fix currently-incorrect solutions
- support automatic append of "WON"
- Remove hack in Text UI that appends ";1" every time
- In Text UI, store SolutionSteps, not text, and implement serialisation of them for printing the output

colonelfazackerley:
I tried the solution to hard 4
```
:solution.1=block;(7,3);7;bridge;(9,13);10;(12,5);6;6;4;4;2;(9,3);dig;(12,10);10;10;10;10;8;explode;(7,2);3;1
```
it's does not work, and it's not just the number of time steps at the end. critical bridge placement time is off.

so, do we have a bug in the text mode recording?

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
```diff
@@ -0,0 +1,128 @@
...
+
+            private SolutionTimeStep waitOneTimeStep()
+            {
+                --waitTimeLeft;
+                return new SolutionTimeStep();
+            }
+
+            private SolutionTimeStep handleAllInstructionsInStep()
+            {
+                ArrayList<Instruction> ret = new ArrayList<Instruction>();
```
tttppp: This is confusing because ret isn't returned. Maybe call it retInstructions?

Also the method doesn't take any parameters, which makes it harder to work out what's going on (Particularly on github where there's no member variable highlighting).

tttppp: I realised afterwards that this is probably deliberate object oriented style, but I still found found less clear than if we had a standalone function.

rabbit-escape-engine/src/rabbitescape/engine/util/Util.java
```diff
@@ -599,4 +599,114 @@ public static boolean isEmpty( String str )
...
+            }
+
+            this.index = index;
+            this.object = object;
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return 31 * ( 31 + index ) + object.hashCode();
```
tttppp: Unusual to have the extra addition of 31*31 in the hash code.

andybalaam:
@colonelfazackerley this code is not yet active,  even in the branch. RunSolution still uses the previous code.  I'll let you know when you can expect bugs to disappear.

colonelfazackerley:
see #161. I am reasonably sure the most egregious bug is in the report that text mode gives when you win.

andybalaam:
Wow, I just discovered `git rebase -i --exec "make clean test" master`

Git is amazing.

andybalaam:
I am literally sitting here with my Rabbit Escape mug, coding and sipping.  Just wanted to share.

andybalaam:
@tttppp:

>This is confusing because ret isn't returned. Maybe call it retInstructions?

Yes, renamed to `instructions`

> Also the method doesn't take any parameters

Yeah I ummed and aahed about that.  I just had a look at changing it, but it's just such a massively stateful method (it modifies `instrs`, `enteredNextStep`, `steps` and `waitTimeLeft`) that I think taking no parameters makes it clear this method is magical.

Any suggestions (or pull requests) to simplify this logic will be gratefully accepted.  The fundamental weirdness is that it translates a list of 2 totally different things (proper Instructions and "magic" WaitInstructions) into a list of consistent things (TimeSteps).  Right now I can't see a way of making it better, although I feel sure there must be one.

> Unusual to have the extra addition of 31*31 in the hash code.

I asked Eclipse to generate a hashCode method, then refactored it into something less egregious.  I see no point in it, and have fixed it.

andybalaam:
Pushed my latest rebase, which will break any checkouts anyone has.

andybalaam:
Merged - it is now active, and doesn't break any tests, so hopefully works as well as before.

PR #163, opened 2015-11-07
Recording simplifier

tttppp:
I think this is significantly better than master, in that the solutions seem to run correctly. However I haven't tried to merge in any of the refactoring that you've been doing on the iterable-solutions branch.


colonelfazackerley:
```
cf@re:~/re/rabbit-escape$ git remote add tttppp https://github.com/tttppp/rabbit-escape.git
cf@re:~/re/rabbit-escape$ git co -b recordingSimplifier
cf@re:~/re/rabbit-escape$ git pull tttppp recordingSimplifier
```
I am on it. I feel like I am getting the hang of this git lark.

colonelfazackerley:
I played hard 10, and the solution just worked. I like the auto-simplification too. Very nice.

I recommend this is merged forthwith. (I will keep using it in my hard-solns branch).

andybalaam:
Merging, but with complaint that there are no unit tests for the new functionality.  Some round-trip tests for the Solutions would be particularly nice.

tttppp:
I didn't think it was worth unit tests because I thought you're refactoring the whole area. I totally agree though that unit tests would improve this!

Thanks for merging anyway, and let me know what you're planning with refactoring this area (i.e. which methods are likely to survive the refactor?)

colonelfazackerley:
If you have something that works, then unit tests are great before a refactor. Easy to see that the new code does the same as the old.

andybalaam:
Refactoring is mostly done, so do go ahead.  I knocked up a few tests here: 072bed55012221a886dd0b397b59e6f2688b9124 which are proving their worth already since quite a few are ignored because they fail (feel free to tell me they are wrong-headed...)

btw I'd prefer the relFormat() code to be outside the SolutionCommand and SolutionAction classes, probably in SolutionParser or similar, with an ActionTypeSwitch, but I'm not demanding you do that.

andybalaam:
Couldn't agree more with @colonelfazackerley 's comment on tests before refactoring.  I'm often surprised how many tests survive even a major refactoring, and they give you so much confidence you've done it right....

PR #164, opened 2015-11-07
Solutions for hard set

colonelfazackerley:
One solution for most. Two solutions for a couple.

I also included @tttppp's fix and simplifier for text mode. I guess that, due to the wonder of git, it wont matter which order you merge them in.


andybalaam:
Superb, thank you.

PR #165, opened 2015-11-08
Until command

tttppp:
This adds support for the until command in text UI and solutions. (No tests yet)


colonelfazackerley:
I will merge this into my hard-solns and arcade branches later, and test it.

colonelfazackerley:
I used until&WON on arcade 1-5 solutions, and that seemed to work.

I tried typing 'until&WON' while playing Ghost versus pie (arcade 1), and got 
```
World.DontStepAfterFinish {cause=null, worldName=Ghost versus pie}
```

tttppp:
Hmm! It's a fail then :-S I only used "until" on its own when playing levels, I tested "until&WON" in solutions. When I was playing using the text UI being able to finish a level with "until" felt pretty good! I won't get a chance to look at the issue for a while sadly.

One other thing I realised was that I don't think I throw an exception if until is still running after 1000 steps. I'm not sure if that's a problem or not.

colonelfazackerley:
Have you handled blockers left over? Other bunnies dead or gone.

tttppp:
Yes - it's actually really simple. It just runs up to 1000 steps, each step checking that the game state is still "RUNNING".

andybalaam:
As implemented here so far, it's "untilWONorWaited1000".  I suggest it should look like `until:WON`, which is a single instruction meaning wait until WON, or `until:LOST` etc.  I'd also like to see unit tests please.

tttppp:
@andybalaam Yes, you're exactly right. I opened a pull request to share this with the Colonel (@colonelfazackerly) in case it made his solution creation easier. It seems that it wasn't the quick win I'd hoped though, so I'm going to close this request and link to it from the enhancement request.

If anyone feels like picking this up, or starting from scratch then just leave a note in the enhancement.

andybalaam:
@tttppp sorry if my crazy refactoring is putting you off.  It's all building on your excellent work, not tearing it down.

tttppp:
@andybalaam Now the branch is finished/merged I'm interested to find out how it works. My concern was really the unknown scope of it. This issue sounds like it should be a really quick win for playability of the text UI, but I must have made a mistake somewhere. Maybe some unit tests will help to locate it.

colonelfazackerley:
Why do you say you made a mistake?

the 'until' you  implemented worked fine for solution playback. its just not compatible with the current master becuase of Andy's changes.

andybalaam:
Yeah I don't think there was anything wrong with it,  but I am!afraid it will be quite different after my changes. If you do it it will give you a chance to get stuck in to the way I've restructured it. I'd be interested to hear whether you like it.

PR #168, opened 2015-11-11
Text ui: falling out of bounds 

colonelfazackerley:
bonus: print solution so far during text mode games. Often the player wants to put a token next to the previous token. instead of squinting and trying to read the column numbers, can calculate relative coordinates.

fixes 149
only thing I am not sure about here: as well as skipping the set() call I am skipping the rabbit.saveState() call. I don't know what that does, and it sounds like a strange thing to be in the rabbit position drawing call.

OutOfBounds does not implement saveState, so I guess its OK.

Tests are passed, and I played a few levels in text ui and all seems well.


andybalaam:
saveState here really means "get state to be saved" but instead we are displaying it not saving it. Pull requests with a better name for that method gratefully accepted, but the point here is that skipping it is fine.

andybalaam:
Superb, thanks!

PR #169, opened 2015-11-12
implement until:WON for solutions in rel files

colonelfazackerley:
Probably not quite ready yet: posting this for advice.

I dont understand runSolution, and I am aware that my change to it is not pretty.

This works for solutions in rel files.

Don't fear the rebase.


andybalaam:
Sorry, the code is misleading - the WaitAction and UntilAction logic should be implemented inside SolutionExecutor. We should be able to delete the body of WaitAction.performAction (and UntilAction.performAction) since they are no longer needed.

andybalaam:
Correction: we should be able to delete the body of caseUntilAction and caseWaitAction.

colonelfazackerley:
I can have a look at moving stuff to SolutionExecutor.

The problem I wanted to raise is in the runSolution method. It seems to want the recorded solution to stop one step short of WON, then step over the line in a seperate step() call. until:WON needs to step all the way to WON. 

So, the until:WON is a special case. but it looks like what was there already was not great, so it may be better to change that in some way.

andybalaam:
SolutionExecutor flips the logic upside-down.  Actions are not responsible for step()ing the world at all - actions get converted by SolutionExecutor into TimeSteps.  A WaitAction or UntilAction translates into multiple TimeSteps.  The problem you are pointing out may still apply, but I can't get my head around it right now.  Update us on what you find...

andybalaam:
(Or ping me if SolutionExecutor just doesn't make sense to you)

colonelfazackerley:
I made a change that looks sensible to me. Now I re-read what you wrote, it does not seem the same.

My problem is with SolutionRunner.runSolution(). it looks wrong to me, but any changes make loads of tests fail. those tests look unimportant.

andybalaam:
I will try and look soon. Hope I haven't been talking rubbish.

andybalaam:
OK, I am talking rubbish.  The way I've done SolutionExecutor, it can't handle an `until` command, unless we start passing the current completionstate in at every step.  I am going to play with that.

andybalaam:
@colonelfazackerley I deliberately removed the performing of actions from the actions themselves.  I wanted them to be pure data that gets parsed into and out of text solution language, without concerning themselves with actually doing things in worlds.  That's why I had a separate doPerformAction method.

andybalaam:
Also, I've just realised I have been saying SolutionExecutor when I meant SolutionInterpreter.  Sorry, this is the consequence of writing a lot of my comments on my phone without access to the code.

andybalaam:
Have to stop now, but I got a basic idea of how to do the until command working in 766e1a19feebacec9216c98af5136df3869f28d3.  Lots to do to flesh it out.

PR #173, opened 2015-11-14
allow comments in rel files. lines starting '%' are ignored

colonelfazackerley:
```
// (C++) used in world ASCII art
/* (C) used in world ASCII art
# (bash) used in world ASCII art
% (GNU Octave alternate comment style) available
```

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/textworld/LineProcessor.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/textworld/LineProcessor.java
@@ -217,6 +217,10 @@ public class LineProcessor
             {
                 processMetaLine( line, variantGen );
             }
+            else if ( line.startsWith( "%" ) )
+            {
+                // Ignore comment
+            }
             else
             {
                 processItemsLine( line, variantGen );
--- a/rabbit-escape-engine/src/rabbitescape/levels/hard/level_04.rel
+++ b/rabbit-escape-engine/src/rabbitescape/levels/hard/level_04.rel
@@ -5,6 +5,7 @@
 :hint1=Find a good place to block so you can bridge to the exit.
 :hint2=The first rabbit should bridge then become a blocker.
 :hint3=Bridge leftwards after climbing the stairs.\nBlock in the middle of the next platform.\nWait, then bridge from the right of that platform as the rabbit turns around.
+% This is not the intended solution. Another is needed
 :solution.1=block;(7,3);bridge;9;(9,13);(12,5);41;(9,3);dig;(12,10);explode;69;(7,2);2;WON
 :num_rabbits=10
 :num_to_save=5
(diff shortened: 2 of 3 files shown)


tttppp:
Great stuff!

andybalaam:
Brilliant - unit tests?  Further thoughts:

- annoying that we have closed avenues for more "normal" comment syntax.  We could use `;` which is a comment in Lisp I think, or `--` which is Haskell and Elm.  I wondered whether `:#` would work, as a special kind of `:` line, avoiding adding to the top-level namespace (which consists of `:` and "everything else") at the moment.  No reason to do any of this - just musing and wondering what you thought.

- would we like to extend this to allow comments anywhere on a line?

- do we want comments to survive a round-trip?  E.g. if we obfuscate a level by round-tripping it, would we still like the comments to be there

colonelfazackerley:
The rel format is simple. I don't see the need to allow comments on the end of the line.

I like the simplicity of a single character, and I don't see a problem with adding to the top level namespace.

I am not a fan of ```;``` because it looks quite like ```:```.

I don't mind the comments being lost. Whatever makes things simpler. (Losing them may complicate test code)

A single ```-``` would be OK. I prefer ```%``` as I am used to using it as a comment marker.

andybalaam:
Fair enough. Write a couple of unit tests and it's in.

colonelfazackerley:
If something is used in some of the levels, then it is in the unit tests...

Apart from that I can't think what to test.

colonelfazackerley:
Taking this view further, we may be able to remove some tests for other things that are in level files.

colonelfazackerley:
Arguing against myself: the unit test Java files can be read as the specification. So all features should be described in them, to the maximum extent possible.

tttppp:
The only rest I can think of for this is that there's no exception when loading a level with a comment. The level string should be as simple as is allowed to try to just test the one thing.

I think the tests that use the real level files are in theory testing that the levels are valid, where as the unit tests should be testing the code behaves as expected. (This is just a long winded way of agreeing with your second point).

Edit: Another potential test (scraping the barrel here) is that a % symbol can be used in e.g. a level name without stripping the rest of the line. Hmm, not sure about this one.

colonelfazackerley:
The test checks
- comments are lost
- % in the name does not change the name.

andybalaam:
> the unit test Java files can be read as the specification

Yes, definitely.  The names of tests should read like descriptions of behaviour.

> there's no exception when loading a level with a comment

Yes, and probably some minimal check that the lines before and after still work?

> the tests that use the real level files are in theory testing that the levels are valid

Yes, and although they are mixed in with our unit tests, they are really "component" or "system" tests that test a great swathe of behaviours, and therefore will fail if one of a large number of parts fail.  E.g. in this case, if the Makefile behaviour of generating "ls.txt" files fails, these tests will fail.

Highly specific unit tests serve to help track down where the problem lies when there is a failure, as well as serving as specification as you mentioned.

> that a % symbol can be used in e.g. a level name without stripping the rest of the line

Excellent test.  I don't mind whether `%` elsewhere is an error, or starts a comment, or is a normal character, and I don't mind changing our minds about this later, but I'd definitely like either a test pinning down the current intended behaviour, or a comment in the tests saying that it is deliberately unspecified.

andybalaam:
Brilliant, thanks!

PR #177, opened 2015-11-16
Wrap hints and descriptions automatically

### LINKED ISSUE #101: Avoid manual wrapping of descriptions and hints
Manually entering \n in the strings is a hassle. We could
1. Get the code displaying the dialog to figure out how long the lines should be, or,
2. Use a script or something to put the \n chars into the files.

colonelfazackerley:
closes #101 

It's nice seeing the lines all the same length in the description and hints. Should make writing the levels less painful.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/util/Util.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/util/Util.java
@@ -280,13 +280,7 @@ public class Util
      */
     public static String wrapToNewline( String s, int maxChar)
     {
-        String[] ss = wrap( s, maxChar );
-        String ret = "";
-        for ( String l: ss)
-        {
-            ret = ret + l + "\n";
-        }
-        return ret;
+        return Util.join( "\n", wrap( s, maxChar ) );
     }
     
     public static String join( String glue, String[] items )
--- a/rabbit-escape-engine/src/rabbitescape/engine/util/Util.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/util/Util.java
@@ -12,6 +12,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import rabbitescape.engine.err.RabbitEscapeException;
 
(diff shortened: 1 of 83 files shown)


andybalaam:
How will this affect Android?

andybalaam:
Longer term,  we should use HTML style labels on Android and Swing,  but I'm happy to go with this for now.

colonelfazackerley:
- I am assuming its "Tannhäuser" that's the ASCII issue. I will change that for a normal 'a'.
- I will look into Util.join
- I will look at Android, but I have not enabled non-play stuff, so I have no idea how to test it.
- re: HTML. The desc is HTML and the hints are not. I think that is your half-job :-p

andybalaam:
Yes, sorry I did mean the ä. Yes, I did do half a job on HTML dialogs.

andybalaam:
To build+debug you need "dev mode" not just untrusted apps.

colonelfazackerley:
is there a VM platform I can use for android? even if its just a GUI?

I will push the ASCII and join thingies in a mo.

colonelfazackerley:
from INSTALL

"There are Eclipse project files, so "File" -> "Import" -> "Existing projects into workspace" should work. Avoid the rabbit-escape-ui-android directory."

do you just edit the android code in a text editor outside the an IDE?

andybalaam:
See the section in that file entitled "Android".  You will need Android Studio https://developer.android.com/sdk/index.html .  If you use a real Android device on Linux, you will need to do the udev magic here: https://developer.android.com/tools/device.html . Alternatively you can install an emulator from inside Android Studio, using the AVD Manager tool.

andybalaam:
Setting up is somewhat complicated - the instructions I linked to are good, but long.

andybalaam:
By VM platform did you mean similar to an emulator? If not I don't think I understand.

colonelfazackerley:
yes, an emulator.

if I do put the package on my phone, what are the dependencies of "make android-debug"?

colonelfazackerley:
I think it would work on android without any changes. I did just push a commit that removes the code that swaps \n for space. Perhaps you could test it? I will have a go, but I would not like to guess how long it will take me to get something working.

andybalaam:
Yeah I'm happy to try it. I forgot we ignored newlines on Android. It should be fine.

andybalaam:
If you follow the instructions on the Android site about how to install Android Studio, `make android-debug` should work.  That builds the package.  To run the package on my phone I usually click the Run button on Android Studio.

To enable developer mode, you don't need a package - just to press the right combination of keys: http://www.androidcentral.com/how-enable-developer-settings-android-42

andybalaam:
Works perfectly on Android.  Nice change, thanks.

PR #179, opened 2015-11-19
bash script to launch swing or text ui

colonelfazackerley:
i would like to combine the mains, so we have a single text main() and a single swing main(). the arg parsing would be done in java. that's the next step.


andybalaam:
Excellent. I suggest `--help` just because it's standard, and I suggest the script be called `rabbitescape` but am open to persuasion on both counts.

andybalaam:
I also agree with your next steps.

colonelfazackerley:
I am fine with --help (and alternatively -h). runrabbit amuses me...

andybalaam:
I thought it might.

PR #180, opened 2015-11-19
start implementing until:WON for solutions

colonelfazackerley:
I started trying to get this working again. I am keen, as I have used until:WON in the arcade set solutions from a while back when it temporarily worked.

The arcade set is ready apart from this.

I got stuck
- I don't understand why UntilAction does not wait, it only asserts.
- I don't understand how this TODO should be done
```
        // TODO: stop using WaitAction to step in the command line interface -
        // then we can stop catching DontStepAfterFinish.
```

It's probably best if you do it.


andybalaam:
I'm sorry I haven't got back to this yet but I will. Hopefully my changes will make it clear what I am intending,  but the fact it is unclear is making me doubt my strategy. **takes self outside for a stern word** No, I am still sure what I want is better. I will do it soon.

andybalaam:
I merged this commit in my recent changes, outside this pull request.

andybalaam:
To explain a bit more: the reason I think my strategy (understand and "interpret" solution actions, translating them into "time steps") is better than the previous one ("run" each solution action, stepping the world as many times as needed for e.g. a wait command) is because I see this being driven in other contexts (e.g. the UIs) where we need to step one time step at a time (e.g. displaying graphics at each time step) so we can't have the actions in control.  Instead, we must invert control and have the UI asking for each time step, and the interpreter able to step one time step at a time.  

Does that in any way help explain what I have been up to?

colonelfazackerley:
Makes sense

andybalaam:
Phew

PR #183, opened 2015-11-20
Outdoors solutions for levels 1, 2 and 3.

tttppp:
I'm not sure when I'll next get chance to work on this, so here are three solutions for outdoors levels.  Saw (level 2) is significantly harder than I remembered - I think this is due to the changes in behaviour since it was written.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/levels/outdoors/level_01.rel
+++ b/rabbit-escape-engine/src/rabbitescape/levels/outdoors/level_01.rel
@@ -3,6 +3,7 @@
 :hint1=Can you climb onto each branch?
 :hint2=Bash through the tree. Wait for the first rabbit to turn around then bridge up to the trunk.
 :hint3=Use a blocker at the end of the rightmost branch.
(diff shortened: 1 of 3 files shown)


andybalaam:
Great,  thanks

andybalaam:
What happened to Saw? Do we need to reorder it?

colonelfazackerley:
I don't think the changes have made Saw harder. Perhaps a case of memory being a funny thing?

andybalaam:
I just tried it and the rabbits no longer enter the exit ( because they fall instead). I think I regard this as a bug in the level. Opinions?

colonelfazackerley:
Yup, I will tweak the level.

I will retract "Perhaps a case of memory being a funny thing".

This is why we need solutions for all levels.

tttppp:
I think the fact that it's still soluble means it's fine, but maybe not quite the level that the author intended.

andybalaam:
> This is why we need solutions for all levels.

Totally agree.

andybalaam:
> I think the fact that it's still soluble means it's fine

I think it's too hard for this order. Candidate for a deja vu level?

PR #186, opened 2015-11-20
Text input

colonelfazackerley:
- token shortcuts (eg d for dig)
- can omit brackets for coordinates

i tried to do this in a way that would not hamper & use or future expansion to allow multiple actions to be entered at once, separated by ';'.


andybalaam:
Cool! Are there unit tests for those regex utils? I found it tricky to work out what they did and tests probably would have helped. Also if we manage to make those system tests this would be a good candidate for them.

andybalaam:
Do me a couple of unit tests for `expandAbbreviations` and I'll happily merge.

colonelfazackerley:
tests done

andybalaam:
Awesome.

PR #187, opened 2015-11-21
Solutions for easy set

colonelfazackerley:
level 1 already has a solution. this is 2-20. 'make test' works.

running through these gave me confidence in the win report from text mode.


andybalaam:
Excellent, thanks.

PR #189, opened 2015-11-22
auto test game for text ui

colonelfazackerley:
expect is weird and twitchy. by fiddling more or less at random, i got this to work.


andybalaam:
Cool . We can expand this to run all levels, check printed solutions match what we typed etc. as we get time.

PR #190, opened 2015-11-23
Text ui: out of bounds bridge building

colonelfazackerley:
- Did not seem right to stop the engine from allowing OOB bridges, because they are used in swing.
- Seemed bad to check bounds for each block as it is drawn, as that sounds like a performance problem

So I added the boolean to the Block class, which is used by the text ui renderer to avoid trying to draw OOB bridges.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/Block.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/Block.java
@@ -17,13 +17,34 @@ public class Block
     public final int y;
     public final Type type;
     public final int variant;
+    public final boolean outOfBounds;
 
+    /**
+     * Blocks that are added later (bridges built by rabbits, for example) 
+     * may be tested for bounds as they are added.
+     */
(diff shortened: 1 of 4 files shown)


andybalaam:
I'd much prefer to bounds check during rendering. I challenge you to demonstrate a statistically significant performance difference :-)

andybalaam:
Also: taste issue - I'd prefer to avoid use of `continue` keyword except where it significantly improves readability. I argue it doesn't do that here.

andybalaam:
Also, thanks for working on this!

colonelfazackerley:
I am working on a new version of this. I have a weird problem with some hamcrest thing.

andybalaam:
If you `static import ...*` from both junit and hamcrest you can experience issues. If this is the problem try importing * from hamcrest and individual members from junit.

colonelfazackerley:
I think you prefer #197

PR #192, opened 2015-11-23
Fix saw

colonelfazackerley:


colonelfazackerley:
It does have an exit on a slope. the animation is not pretty.

andybalaam:
Brilliant, thanks. This level already exposed loads of animation bugs, so 1 more is fine.

Issue #159, opened 2015-11-07
Remove Ant from the build process


andybalaam:
Speed up builds by not using Ant.

We only use Ant to do `javac` (which is very dumb and doesn't know anything about inter-file dependencies, so we can do the same in Make) and `junit`, which can be replaced with a bit of grep or find.

andybalaam:
I committed a change that uses a bash script instead of Ant to compile: ee6b7df61240c85a9d256ea7bda1d89a9302f510

andybalaam:
Notes to self:
* make compile and test targets should use touchfiles to speed up
* generic jar contains TestClasses if you build in the default order

andybalaam:
Removing Ant for tests is here: 0cdb8034

andybalaam:
Using touchfiles is here: 5449f13f

PR #194, opened 2015-11-26
For consideration: compiling without Ant

andybalaam:
This is for discussion, not merging now: it's not ready to go yet - files get compiled into different places, it breaks Eclipse etc.
## Performance

The current code (with Ant) performs like this:

```
$ make clean && time make compile
16s

$ time make compile
2.6s

$ touch rabbit-escape-engine/src/rabbitescape/engine/ChangeDescription.java 
$ time make compile
4s
```

With this pull request:

```
$ make clean && time make compile
12s

$ time make compile
0.9s

$ touch rabbit-escape-engine/src/rabbitescape/engine/ChangeDescription.java 
$ time make compile
2.2s
```

So there is a noticeable performance improvement, especially when not building from clean.
## Arguments
- Faster is better
- This pull request gives us a more complex Makefile (I am thinking maybe we need to split up our Makefile btw)
- This pull request helps us ditch a dependency on Ant, which reduces complexity
## Your thoughts

Should we pursue this?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,22 +199,34 @@ no-awt-in-engine:
 	! find rabbit-escape-engine/src -name "*.java" -print0 | xargs -0 grep 'java\.awt'
 	! find rabbit-escape-render/src -name "*.java" -print0 | xargs -0 grep 'java\.awt'
 
+
+
+JAVA_DIRS := rabbit-escape-engine rabbit-escape-render rabbit-escape-ui-swing rabbit-escape-ui-text
+JAVA_FILES := $(shell find ${JAVA_DIRS} -name "*.java")
+CLASSES_DIR := bin/classes
+CLASSES_MARKER := bin/classes.marker
+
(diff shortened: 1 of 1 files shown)


andybalaam:
@colonelfazackerley @tttppp would appreciate your thoughts.

andybalaam:
Related to #159

colonelfazackerley:
I am not particularly good with Make and I know nothing of Ant, so I am going to keep quiet.

tttppp:
I would tend to go with simpler code, longer compile time (provided we're talking about a few seconds).  I'm not sure my opinion's going to count for much here though, as I'm used to using maven!

andybalaam:
Removing ant makes life simpler though. I've been using maven recently. Nightmare.

andybalaam:
I still want to do this, but leaving this PR open is not helping.  We can still refer back to it if/when we do it.

PR #195, opened 2015-11-26
Tests without Ant

andybalaam:
This one seems pretty clear cut.  The change is quite simple, and the performance gain is decent.

Current code with Ant:

```
$ time make test
9.0s
```

Without Ant (this pull request):

```
$ time make test
6.2s
```

It's less code, with just one nasty line of sed, that can probably be simplified and/or replaced with Makefile magic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,9 +260,14 @@ runlevel: compile
 runat: compile
 	java -cp $(CLASSPATH) rabbitescape.ui.swing.AnimationTester
 
+JAVA_DIRS = rabbit-escape-engine rabbit-escape-render rabbit-escape-ui-text rabbit-escape-ui-swing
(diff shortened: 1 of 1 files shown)


andybalaam:
Again, comments welcome @tttppp @colonelfazackerley

andybalaam:
Related to #159

colonelfazackerley:
I don't understand what Ant does for tests, so I can't really comment.

andybalaam:
Ant searches all java code for the `@Test` annotation I think, whereas we can find all files named Test* because we happen to know all our tests are named that way.

andybalaam:
I still want to do this, but leaving this PR open is not helping.  We can still refer back to it if/when we do it.

PR #197, opened 2015-11-27
Text oob bridge 2

colonelfazackerley:
not working. seemingly unrelated things are failing tests. hoping someone can tell me why.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/textworld/Chars.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/textworld/Chars.java
@@ -56,11 +56,14 @@ public class Chars
     private final boolean starsMode;
     private final char[][] impl;
     private final Map<Point, String> stars;
+    private final int worldWidth, worldHeight;
 
     public Chars( World world, boolean starsMode )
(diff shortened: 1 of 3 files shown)


colonelfazackerley:
something that I have done to Chars.java is breaking All_levels_load_and_round_trip

colonelfazackerley:
Stupid mistake. I think I have it.

colonelfazackerley:
Stupidity rebased away. Please merge now.

andybalaam:
Great.

colonelfazackerley:
Would it have been terrible to catch and ignore the OutOfBoundsException instead of checking bounds in the code? the JVM does bounds checking anyway...

andybalaam:
It would have been good Python style.  In Java we normally avoid this, but not necessarily for good reasons.  In C++ throwing exceptions is significantly slower than normal operations, because/hence exceptions are considered only to be used in exceptional circumstances.

Personally I find the syntax of exceptions (in Java) horrible, so I try to avoid them where possible, using them only to bail out through several levels of code because something went wrong.

PR #201, opened 2015-11-28
Outdoors solutions for levels 4 to 8.

tttppp:
Level 8 contains the "save all 4 solution", rather than the author's
intended solution (which I don't know).


andybalaam:
Fantastic, thank you!

PR #202, opened 2015-11-28
Top slope bash

### LINKED ISSUE #153: Bug: rabbit at top of slope should bash uselessly
```
rb/
#/#
   
rb 
#/#
```
The top rabbit bashes out the next slope block. The lower rabbit bashes out the block behind the slope block he is on, which looks wrong.

I think the lower rabbit should consume the token and then levitate half a square (like the upper rabbit) and bash in a futile manner against air. Some discussion is needed, as I have some uncertainty about this opinion.

I wondered if this one is my fault, so I did "git checkout 0a35". This behaviour has been there for a while.

colonelfazackerley:
fixes #153


andybalaam:
Very cool, thanks.

andybalaam:
Animation leaves a little to be desired (as in many cases, including non-useless bashing on a slope) but the mechanics are now right.

PR #208, opened 2015-11-30
display level name in ...swing.TopBar

colonelfazackerley:
i was going to try and do the level names in menus too, but that may be a while away, so here's this simpler change on it's own.


andybalaam:
Excellent,  thanks!

PR #210, opened 2015-12-02
animation for dying of falling onto slope rising right: 7 custom frames

colonelfazackerley:
Advice on the animation please.


colonelfazackerley:
I think this could be pulled now.

needs another 7 frames for slopes rising left and another 3 states, I think. I will do those after some feedback.

oh, and a level to see the animation.

andybalaam:
Brilliant. I will look when I get on the computer.

colonelfazackerley:
fall

The screenshot recorder I used (byzanz) makes this look a bit glitchier than it is.

tttppp:
I like this! More animations presumably means more Android crashes?

andybalaam:
Really cool!  @tttppp yes. I will work on those crashes soon. Basic plan: 1. Catch OutOfMemory errors and shrink cache.  2. Stop holding 2 copies of each image (raw and resized). 1. must come first since 2 doesn't prevent crashes, just makes them harder to reproduce.

andybalaam:
I am thinking we should leave this until after I've made a release containing the arcade levels. If it not being merged is causing a problem we can make a branch.

colonelfazackerley:
If you try it and merge to master, I can probably get rest done in about 3 days. then it can go in the next release. i think this is what needs doing to finish it off
- another rea file for falling 2 squares and splatting in one time step onto a right rise. this can use the same frames
- 7 frames for splatting on a left rise
- 2 more rea files for splatting on a left rise.
- integration of the new states (rea files) into SHOUTY_CAPS in the code.

andybalaam:
I'd prefer to isolate any further code changes from this release just because I've tested it as it is now and am happy with it. Of course if I take ages over doing the release this will become annoying and we should change our minds.

colonelfazackerley:
So the animation and state changes here are OK?

I will finish off the list above.

andybalaam:
Yeah it looks great!

PR #213, opened 2015-12-06
Fix arcade meta

colonelfazackerley:
The only outstanding item from the bits @andybalaam has mentioned is a new description without the bash/bridge behaviour switch for Space invaders.

Perhaps someone else fancies that job?


andybalaam:
I can do it if no-one gets to it first. Thanks again for all your work on this.

PR #215, opened 2015-12-08
Set dir name

colonelfazackerley:
I think this will be a necessity when we have 10 sets.

note the bonus commit is because of a dumb fall-through warning I get sometimes.


andybalaam:
I am fine with this but don't understand the need for it?

colonelfazackerley:
When looking at
https://github.com/andybalaam/rabbit-escape/tree/master/rabbit-escape-engine/src/rabbitescape/levels
the sets are not in order. You can ignore my OCD if you like.

andybalaam:
Ocd +1

andybalaam:
D'oh, should probably keep this until after release too.  Hopefully pretty low-risk.

colonelfazackerley:
'make test' is fine.

I also opened a level from each set in swing. I can't see what kind of cock-up could make anything else go wrong.

andybalaam:
Yes.  I will obviously sanity check on Android before release too.

andybalaam:
And run the smoke test.  It will be nice when we have a system test that plays all levels on each platform too...

andybalaam:
(using the provided solutions)

andybalaam:
I've just realised this creates an upgrade issue because the config that remembers what levels you have completed uses the dir name.

We are going to have to support upgrading config over releases, so this is fine, but in order to apply this change we need an upgrade.  I've had to revert it (1a1c3d8d8b835b03f06535233f19355b643252cc) so I can get release 0.7 out.

Sorry for messing you about.

The upgrade will need to check whether we have old config and modify the level dirs to the new names.  We should introduce a config version numbering system so it's easy to detect what version you have.  This upgrade will need to work on Android as well as desktop.

andybalaam:
This demonstrates why I should not have merged this until after the release :-)

PR #219, opened 2015-12-12
dying on slopes continued

colonelfazackerley:
I set this up so you can see progress. (not ready yet)


colonelfazackerley:
now I have most of the rabbit melting away, then the head rolls down the hill and one of the eyes comes out.

mwahahahaha

(needs integrating into the game, but looks ok in AT)

colonelfazackerley:
now I am back to where I thought I was.

colonelfazackerley:
all the frames are done now. just need 2 more rea files for falling even numbers. integrate that into states. fix tests.

on left slopes the head starts rolling and one of the ears flies off.

colonelfazackerley:
ready to be pulled!

colonelfazackerley:
slope-splat

andybalaam:
Beautiful! One thing to check: I thought I noticed some of the lines getting thin (probably where you shrunk shapes). I often find I have to adjust line widths after shrinking shapes. Normally all lines should be the same width no matter what size the shape is.

colonelfazackerley:
I noticed that. I thought it was OK in the case of a death animation. Various body parts need to disappear. Having the lines thin down helps them to shrink to nothing gradually.

andybalaam:
I'll look when I can get on a computer, but I am likely not to like it since it breaks the cartoon style.

andybalaam:
Fading out has been used before,  which looks ok.

colonelfazackerley:
OK. I might try thickening up the lines, but fading them.

PR #231, opened 2015-12-26
enable exceptions for levels with no solution that are referenced by …

### LINKED ISSUE #182: Solutions for all published levels should be mandatory
Currently, if you have a level with no solution, `make test` passes, but it should fail.

This may be annoying while we are developing levels - suggestions welcome.

colonelfazackerley:
…the menus. closes #182


tttppp:
Good spot :-)

andybalaam:
Nice, ta

PR #232, opened 2015-12-26
Test anim

colonelfazackerley:


andybalaam:
Yes, very good, thanks.  Some whitespace needs fixing btw.

andybalaam:
btw do you hate the changes I have been making to make the build quieter?  I wish it could be verbose when it failed and quiet when it succeeded.

colonelfazackerley:
I mostly prefer it. when it fails, I can't see why, though. I mostly use 'make test', and when it fails I can now do 'make test-verbose'

andybalaam:
I guess we have to live with that until we find something better

PR #251, opened 2016-01-02
record solutions in swing

### LINKED ISSUE #250: Record solutions in swing
I am working on this.

colonelfazackerley:
closes #250 

this works now and could be pulled. at the moment it prints to System.out whenever a level is completed

* should it only print based only command line options?
* only print based on config (which the user would have to enable by manually editing the file)
* pop up a dialog with the solution? (yuck)
* should the report include ":solution.1="?

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
@@ -12,7 +12,14 @@ public class SolutionCommand
 
     public SolutionCommand( CommandAction... actions )
     {
-        this.actions = actions;
+        if ( 0 == actions.length )
+        {
+            this.actions = new CommandAction[] { new WaitAction( 1 ) };
+        }
+        else
+        {
+            this.actions = actions;
+        }
     }
 
     @Override
@@ -33,6 +40,32 @@ public class SolutionCommand
         SolutionCommand otherSolution = (SolutionCommand)other;
 
         return Arrays.deepEquals( actions, otherSolution.actions );
+    }
+    
+    /**
+     * Try to combine two commands. If this is not possible then return null.
+     */
+    public static SolutionCommand tryToSimplify(
+        SolutionCommand existingCmd, SolutionCommand newCmd )
+    {
+        if( null == existingCmd || null == newCmd )
+        {
+            return null;
+        }
+        CommandAction action1 = existingCmd.actions[0];
+        CommandAction action2 = newCmd.actions[0];
+
+        if (
+               action1 instanceof WaitAction
+            && action2 instanceof WaitAction
+        )
+        {
+            WaitAction wait1 = (WaitAction)action1;
+            WaitAction wait2 = (WaitAction)action2;
+            return new SolutionCommand(
+                new WaitAction( wait1.steps + wait2.steps ) );
+        }
+        return null;
     }
 
     @Override
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
@@ -10,6 +10,11 @@ public class SolutionCommand
 {
     public final CommandAction[] actions;
 
+    /**
+     * If the supplied array (vararg) of CommandActions is empty
+     * a default WaitAction of 1 step will be created. The solution fragment 
+     * ";;" has implied ones, "1;1;".
+     */
     public SolutionCommand( CommandAction... actions )
     {
         if ( 0 == actions.length )
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
@@ -3,7 +3,7 @@ package rabbitescape.engine.solution;
 import java.util.ArrayList;
 import java.util.List;
 
-public class SolutionRecorder
+public class SolutionRecorder implements SolutionRecorderTemplate
 {
     private List<CommandAction> commandInProgress;
     private final List<SolutionCommand> solutionInProgress;
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
@@ -6,7 +6,7 @@ import java.util.List;
 public class SolutionRecorder
 {
     private List<CommandAction> commandInProgress;
-    private List<SolutionCommand> solutionInProgress;
+    private final List<SolutionCommand> solutionInProgress;
     
     public SolutionRecorder()
     {
@@ -19,11 +19,8 @@ public class SolutionRecorder
         commandInProgress.add( a );
     }
     
-    public void appendStepEnd()
+    public void append( SolutionCommand newCmd )
     {
-        CommandAction[] aA = new CommandAction[commandInProgress.size()];
-        aA = commandInProgress.toArray( aA );
-        SolutionCommand newCmd = new SolutionCommand( aA );
         int prevCmdIndex = solutionInProgress.size() - 1;
         SolutionCommand prevCmd =   prevCmdIndex >= 0 
                                   ? solutionInProgress.get( prevCmdIndex )
@@ -37,7 +34,21 @@ public class SolutionRecorder
         { // Replace previous with combined
             solutionInProgress.set( prevCmdIndex, combCmd );
         }
-        
+    }
+    
+    public void append( Solution solution )
+    {
+        for ( SolutionCommand command : solution.commands )
+        {
+            append( command );
+        }
+    }
+    
+    public void appendStepEnd()
+    {
+        CommandAction[] aA = new CommandAction[commandInProgress.size()];
+        aA = commandInProgress.toArray( aA );
+        append( new SolutionCommand( aA ) );
         // Prepare to collect actions in the next step.
         commandInProgress = new ArrayList<CommandAction>();
     }
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
@@ -1,0 +1,37 @@
+package rabbitescape.engine.solution;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SolutionRecorder
+{
+    private List<CommandAction> commandInProgress;
+    private List<SolutionCommand> solutionInProgress;
+    
+    public SolutionRecorder()
+    {
+        commandInProgress = new ArrayList<CommandAction>();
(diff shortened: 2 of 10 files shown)


colonelfazackerley:
First time i tried this I recorded the ";" from GameLoop.step and get ten times more than intended. Hohoho.

andybalaam:
This is (obviously) a good idea and I will get to looking at it at some point.

andybalaam:
If/when I make a proto-level editor, the UI for this will clearly live in there.

andybalaam:
I think `SolutionRecorder.append` should take in a `CommandAction` and there should be an additional method called `SolutionRecorder.nextTimeStep()` (or indeed just `step()`).  Inside, the SolutionRecorder can just build up a `Solution` and use the existing serialisation code to print out the answer.  This concatentating strings lark is no fun.

(Even better, but more work: `append` should take in a `TimeStepAction`, and then we write a backwards translator that can convert a list of `SolutionTimeStep` back to a `Solution`.  This is the reverse of what the `SolutionInterpreter` does.)

colonelfazackerley:
How about this? I rebased away the stringy implementation.

I see there are conflicts now. Should I 
```
git fetch upstream
git co master
git merge upstream/master
git co rec-swing-solns
git merge master
# resolve conflicts
# commit the resolution
# rebase away the merge
```

colonelfazackerley:
or, is this the same, but more concise
```
git co rec-swing-solns
git pull --rebase upstream master
```
and my master, is fine, because i never commit to it, so I can do
``
git co master
git pull upstream master
```

**Comment by andybalaam** [comment]
Yes, your last comment is correct.

**Comment by andybalaam** [comment]
I tend to `git checkout -b tmpbranch` before I start so I have a branch I can return to if everything goes bad. Of course `git rebase --abort` should save you but...

**Comment by colonelfazackerley** [comment]
rebased away the (very minor) merge issue. ready

File: rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java [hunk 1]
```diff
@@ -12,7 +12,14 @@
...
-        this.actions = actions;
+        if ( 0 == actions.length )
+        {
+            this.actions = new CommandAction[] { new WaitAction( 1 ) };
+        }
+        else
+        {
+            this.actions = actions;
+        }
     }
```
andybalaam: Why are we adding a wait if you didn't pass anything in?

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
```diff
@@ -0,0 +1,51 @@
...
+            solutionInProgress.add( newCmd );
+        }
+        else
+        { // Replace previous with combined
+            solutionInProgress.set( prevCmdIndex, combCmd );
+        }
+        
+        // Prepare to collect actions in the next step.
+        commandInProgress = new ArrayList<CommandAction>();
+    }
```
andybalaam: Is there similar logic in InputHandler that can be combined?

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
```diff
@@ -18,15 +19,21 @@
...
         {
             this.world = world;
+            this.solutionRecorder = solutionRecorder;
         }
 
         public synchronized void step()
         {
             world.step();
+            if ( null != solutionRecorder )
+            {
```
andybalaam: Why allow this to be null? I generally prefer not to.

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGameLaunch.java
```diff
@@ -381,6 +387,7 @@ public int addToken( int tileX, int tileY, Token.Type ability )
         {
             graphics.playSound( "place_token" );
         }
+        solutionRecorder.append( new PlaceTokenAction( tileX, tileY ) );
         return now;
```
andybalaam: Weird that we append in both GameUI and here. Might be symptomatic of wider bad design or might be fixable?

andybalaam:
Liking it overall btw. Next up,  swing solution playback?

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
colonelfazackerley: if time steps but nothing has been done the default is a wait of 1. like just hitting enter and typing nothing in the text UI. Made sense to me.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionRecorder.java
colonelfazackerley: Quite possibly. Some stuff should move from the text package to SolutionRecorder. I will look at that.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
colonelfazackerley: I was thinking that we may not always want to be recording solutions. Is this code used on android, where we are not recording solutions, for example?

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGameLaunch.java
colonelfazackerley: The append in GameUI is switching token type, and the append here is placing a token. I think this is fine and correct.

colonelfazackerley:
yep, that would be cool. not too hard ... possibly ...

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
andybalaam: Couldn't we pass on an array of size 1? I don't feel like that knowledge should live here.

andybalaam: In, not on sorry.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
andybalaam: I guess I'd prefer to pass in an inactive recorder rather than have null checks everywhere.

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGameLaunch.java
andybalaam: But why are these 2 actions happening in different classes? I don't think it's a fault in your code but I am curious.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
colonelfazackerley: I did something like that at first. But null actions don't exist, for example ';;;;' has implied ones like '1;1;1;1;', so this felt right. I can do it the other way if you insist.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
colonelfazackerley: OK, I can try that. I still can't test Android.

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGameLaunch.java
colonelfazackerley: I followed where the mouse clicks went, and inserted this into the business end. I guess it's because ability selection is a mouse button click and that is handled differently to a token drop where you need to find out where on the main game panel a click has happened and map it to a location in the game's world.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
andybalaam: Hmm. I see what you mean. Cool. Just a comment to that effect them please.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
andybalaam: I am soon going ya make it possible/mandatory to compile Android with each build.

andybalaam: Sorry for my stupid phone typing.

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGameLaunch.java
andybalaam: Right. Not your fault but one to think about.

andybalaam: (Later)

colonelfazackerley:
how about this? i think i have done everything discussed (apart from swing demo mode, but that's another pull request).

andybalaam:
Excellent.

andybalaam:
Musings naming of `SolutionRecorder`:
1. Could we just pass in a real `SolutionRecorder` even when we don't care about the solution? 
2. I'd argue that the interface should be called `SolutionRecorder` and the empty implementation could be `IgnoringSolutionRecorder`. I admit it's hard to think of a good name for the real one then.

colonelfazackerley:
AttentiveSolutionRecorder perhaps.

colonelfazackerley:
I prefer the names with Solution first.

andybalaam:
I certainly reject the java convention of cramming more and more names on the front as we derive more and more.

colonelfazackerley:
I suggested Attentive as an antonym for your Ignoring. I prefer Template (or something) on the interface.

I think "Solution" can be removed from all the classes (apart from Solution which would be a null string ...). This would leave Recorder, Parser, Runner, etc. They are in the solution package, after all.

I also think CommandAction should just be Action.

tttppp:
I'd leave Solution on the front if it avoids ambiguity across the project. There aren't that many classes in the game that we need to resort to package names to distinguish between them.

Maybe InMemorySolutionRecorder?

andybalaam:
Command Action is to distinguish from TimestepAction.

colonelfazackerley:
Should I do another pull request removing SolutionIgnorer and SolutionTemplate, and give SolutionRecorder a lazy mode instead?

andybalaam:
What would a lazy mode do? Could we just pass on a real Solution Recorder but not use what it records?

colonelfazackerley:
I was thinking lazy mode would not record. I suppose discarding the record would be the same.

andybalaam:
I like the simplicity of just passing in a recorder whose output we don't use, but if you don't feel inspired to do it I am happy with it as it is.

andybalaam:
Or we could compromise and add a TODO or issue that I will do myself one day if I feel that strongly about it.

PR #266, opened 2016-01-06
Add intended solution for Arcade 13.

tttppp:
The level was meant to be a straightforward race against the traitor, but
the other solution is very clever!


andybalaam:
Should the hints be updated?

tttppp:
I don't really mind. Both solutions are clever in different ways. I did think about adding an extra dig token to the traitor's pile, to circumvent the "cheating" Solution, but in the end I thought that both were really good (and moreover hard to find!)

andybalaam:
I'll have a look at your solution when I get a chance.

PR #270, opened 2016-01-07
Remove text arg from ./runrabbit (now it's the default)

colonelfazackerley:
and some expect tests.

I have not done expect tests for things that produce or change files ("--encode foo.rel" etc. I could not decide how best to tidy up the temp files that result, or where to put them while the test runs)

I still need to change the .bat

Comments so far gratefully received.

### DIFF
--- a/rabbit-escape-ui-text/src/rabbitescape/ui/text/TextMain.java
+++ b/rabbit-escape-ui-text/src/rabbitescape/ui/text/TextMain.java
@@ -62,7 +62,7 @@ public class TextMain
             }
             if ( level.isPresent() )
             {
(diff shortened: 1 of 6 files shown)


colonelfazackerley:
I tested the bat. Note that windows %* does not need quotes like $@ for bash. It works for paths with spaces.

I think this is ready.

colonelfazackerley:
I could not get ```--encode <string>``` to work with expect because of the funny characters produced. I could not find how to quote or escape them so it worked.

colonelfazackerley:
I will do the wiki thing about comments after this. I wanted to mention the command line parameters in that article, as that's what moves the comments around.

andybalaam:
Nice work.

PR #271, opened 2016-01-08
Behave bugs

colonelfazackerley:
one bug down, but don't pull yet, as I might have the another swatted soon.

and I should add a test.


colonelfazackerley:
needs test for second bugfix. almost ready.

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
```diff
@@ -63,6 +63,11 @@ public boolean rabbitIsFalling()
         }
     }
 
+    public boolean rabbitIsClimbing()
+    {
+        return rabbit.state.toString().contains( "CLIMB" );
+    }
+
```
andybalaam: I'd prefer a switch on enum values here. But I think we are seeing more and more cases where we need more structure than a flat list. We should start by removing the redundant left/right repetition.

rabbit-escape-engine/src/rabbitescape/engine/behaviours/Walking.java
```diff
@@ -274,6 +274,7 @@ public boolean behave( World world, Rabbit rabbit, State state )
                 return true;
             }
             case RABBIT_TURNING_LEFT_TO_RIGHT:
+                rabbit.onSlope = false; // Intentional fall-through
             case RABBIT_TURNING_LEFT_TO_RIGHT_RISING:
```
andybalaam: Seems like onSlope is not well-defined enough if we are having to do random fixups like this... Maybe onSlope should become false at the beginning of each time step.

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
colonelfazackerley: OK, I will change that for a switch.

I think using classes with inheritance instead of an enum would make sense. so all climbing clases descend from ClimbingState etc. I guess the parent of RabbitState would have a left/right boolean or something.

Having said that, I don't think I can be bothered with making that change...

rabbit-escape-engine/src/rabbitescape/engine/behaviours/Walking.java
colonelfazackerley: onSlope has been implicated in several bugs. 

Is there some problem with bridge building if it is reset each step without a lot of care?

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
andybalaam: No, and I am suspicious of inheritance as a solution for any problem so would want to explore other options and/or see a prototype.

rabbit-escape-engine/src/rabbitescape/engine/behaviours/Walking.java
andybalaam: Possibly. Again, this is not for this change, just musing. Fortunately we've got good test coverage so can contemplate refactoring.

andybalaam:
Happy to merge when it's a switch.

colonelfazackerley:
done!

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
colonelfazackerley: It seems that the tree structure you can get from inheritance seems a natural way to store all those states needed to map to animations. I am interested to see how else you do it.

andybalaam:
I'm almost beginning to become complacent about your commitment to improving this game detail by detail. Thanks.

PR #276, opened 2016-01-10
swing solution player

colonelfazackerley:
I put this up so progress can be tracked.

Let me know if the class should be renamed SolutionPlayer.

early feedback is valued.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/Player.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/Player.java
@@ -1,0 +1,64 @@
+package rabbitescape.engine.solution;
+
+import java.util.Iterator;
+import java.util.List;
+
+import rabbitescape.engine.util.Util;
+
+public class Player
+{
+    
+    final Iterator<SolutionCommand> commands;
+    private SolutionCommand currentCommand;
+    private int waiting;
+    boolean played = false;
+    
+    public Player( Solution s )
+    {
+        List<SolutionCommand> cmdList = Util.list( s.commands );
+        commands = cmdList.iterator();
+        waiting = 0;
+    }
+
+    public boolean hasMoreSteps()
+    {
+        return waiting > 0 || commands.hasNext() ;
+    }
+    
+    /**
+     * 
+     * @return true if there are non-wait actions this step.
+     */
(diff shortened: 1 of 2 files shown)


andybalaam:
On naming I appreciate the desire to avoid sticking words together but think Player is very generic. How about Playback?

andybalaam:
Most of the logic you've written I think is a duplicate of SolutionRunner - that takes in a Solution and provides an iterator through time steps. That is exactly what you need I think. (And was at the root of why I wanted to refactor the solution stuff I did.

colonelfazackerley:
I can't see how to re-use that code. SolutionRunner seems to need to have it's own copy of the world in a sandbox game. It wants to step the world. For playback in the realtime UIs, I think the realtime UI has to step the world.

Most of it's methods are private: nothing for swing to call to find out what actions to take. 

So, I am going to need some pointers on how to get started with using SolutionRunner for this purpose.

colonelfazackerley:
I create a SandBox game which clones the world used by swing. I then use SolutionRunner.runPartialSolution one command at a time? I can't see a way SolutionRunner is going to tell me for a particular timestep what token selections and drops to make (without more - possibly fiddly - code)

andybalaam:
Ah sorry not SolutionRunner, SolutionInterpreter.

andybalaam:
If it doesn't make sense let's have a chat or Skype about it. It may not for add well as I hoped it would

colonelfazackerley:
SolutionInterpreter makes a lot of sense.

Those ignorant* of the codebase are doomed to attempt to recode it.

Should SolutionRunner move to text?

*me

colonelfazackerley:
Ah, SolutionRunner is used by tests so it belongs in engine.

i think I will kill this branch, and try again in a new pull request.

andybalaam:
Sorry I know my communication around that whole refactoring was horrible. I am glad this has led us to clear it up a bit. When you start to use it hopefully you can recommend how to make it easier to understand too.

colonelfazackerley:
I was probably just being dense. SolutionInterpreter is clear. Naming is tricky. They could probably all do with a little doxygen to explain the difference between runner and interpreter, etc.

andybalaam:
Yes

PR #278, opened 2016-01-10
make wait actions in multi-action commands illegal

### LINKED ISSUE #277: Discussion: multi-action commands including waits should be illegal
I was implementing an exception to be thrown when a solution containing this is parsed.

I found a test with this in "bash&(1,1)&3;WON".

I think this should be illegal. The point of '&' is to allow multiple actions in a command in a single time step, so combining it with a wait defeats its purpose. We agreed '&' should be used as little as possible. 

... and (here it comes) it makes solution playback code fiddlier.

Can I make it illegal and implement that exception?

colonelfazackerley:
I realised when getting the tests to pass that this is not just an oversight.

I think '&' should be used as little as possible.

closes #277


rabbit-escape-engine/src/rabbitescape/engine/err/exceptions_en.properties
```diff
@@ -17,6 +17,7 @@ solution.SolutionExceptions.UsedMissingAbility=Solution failed: ability '${abili
 solution.SolutionExceptions.PlacedTokenOutsideWorld=Solution failed: placed a token at (${x}, ${y}) but the world is only ${worldWidth}x${worldHeight} in size at command ${commandIndex} of solution ${solutionId} in ${level}:\n${world}.
 solution.SolutionExceptions.FailedToPlaceToken=Solution failed: tried to place a ${type} token at (${x}, ${y}) but a block was already there so it did not place at command ${commandIndex} of solution ${solutionId} in ${level}:\n${world}.
 solution.SolutionExceptions.UntilActionNeverEnded=Solution failed: the level never finished, but there was an until:${expectedState} action at command ${commandIndex} of solution ${solutionId} in ${level}:\n${world}.
+solution.SolutionCommand.WaitActionIllegalInMultiActionCommand=Wait actions are illegal in mult-action commands: '${command}'
 util.CommandLineOption.UnknownOption=Unkown option: '${unknownArg}'.
```
andybalaam: Missing i in multi.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionCommand.java
```diff
@@ -4,10 +4,22 @@
 
 import java.util.Arrays;
 
+import rabbitescape.engine.err.RabbitEscapeException;
 import rabbitescape.engine.util.Util;
 
 public class SolutionCommand
 {
+    public final class WaitActionIllegalInMultiActionCommand extends RabbitEscapeException
+    {
```
andybalaam: Should be called WaitActionInMultiActionCommand. The fact that it is illegal is not part of what went wrong.

rabbit-escape-engine/test/rabbitescape/engine/solution/TestSolutionCommand.java
```diff
@@ -13,8 +13,8 @@
     @Test
     public void Equal_solution_commands_are_equal()
     {
-        SolutionCommand solution1 = makeSolutionCommand( 3 );
-        SolutionCommand solution2 = makeSolutionCommand( 3 );
+        SolutionCommand solution1 = makeSolutionCommand();
+        SolutionCommand solution2 = makeSolutionCommand();
 
         assertThat( solution1, equalTo( solution2 ) );
```
andybalaam: Should add a similar tests for commands that only contain waits if they don't already exist

andybalaam:
Looks good. Will merge with those comments addressed or answered.

colonelfazackerley:
I think I addressed all your points.

andybalaam:
This makes us better. My tests for these cases were a classic example of choosing the complicated solution over the better one.

PR #279, opened 2016-01-11
swing solution player attempt 2

colonelfazackerley:
Not working yet. Posting this so I can be told I am doing it wrong nice and early in the process :)

NothingPlaying is a slightly different way of getting round ugly null iffing in Physics.


rabbit-escape-engine/src/rabbitescape/engine/solution/NothingPlaying.java
```diff
@@ -0,0 +1,17 @@
...
+    {
+        super( SolutionParser.parse("1") );
+    }
+    
+    @Override
+    public SolutionTimeStep next( CompletionState worldState )
+    {
+        return new SolutionTimeStep( ++commandIndex, new TimeStepAction[] {} );
+    }
+}
```
andybalaam: Can SolutionInterpreter just take a flag to tell it to emit empty time steps when it's past the end?  I don't like deriving from concrete classes where it can be avoided.

andybalaam: Or use the interface Iterator<Timestep> instead of taking in a SolutionInterpreter and make NothingPlaying just implement that. 

But I think you'll want SolutionInterpreter to keep emitting empty time steps here anyway (after the solution is finished).

andybalaam: I'm talking rubbish - you'd need to define am interface and make SolutionInterpreter implement it.

colonelfazackerley:
I made the change you suggested.

Completed basic implementation. It works! (At least for 1 level).

I am sure it needs some tweaks.

colonelfazackerley:
playing in normal swing campaign mode seems OK.

colonelfazackerley:
clicking while a solution is running can make you run out of tokens, and the solution fails. i think this is correct.

colonelfazackerley:
a solution that runs out of steps throws a null pointer exception. I think the interpreter should just go quiet and let the level run.

colonelfazackerley:
I can see a few ways of dealing with this, but I am not sure which will be your preference.

colonelfazackerley:
a solution with an excessive number of waits on the end is fine.

colonelfazackerley:
if the solution has tokens placed illegally on a block square, it is ignored. i think it should probably stop with an exception.

tricky, because I am using the swing UI code, which ignores clicks in the wrong place.

colonelfazackerley:
excessive token drops in the solution are quietly ignored. perhaps this is correct, as the drops could be a mix of user clicks and solution commands.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
```diff
@@ -104,6 +124,25 @@ public boolean gameRunning()
...
+        for ( TimeStepAction action: stp.actions )
+        {
+            if ( action instanceof SelectAction )
+            {
+                actionTypeSwitch.caseSelectAction( (SelectAction)action );
+            }
+            else if ( action instanceof PlaceTokenAction )
+            {
+                actionTypeSwitch.casePlaceTokenAction( (PlaceTokenAction)action );
+            }
```
andybalaam: Not sure reusing TimeStepActionTypeSwitch here is very clear? I can live with it if you defend it but I was expecting these type switch classes to only be implemented inline.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
```diff
@@ -45,9 +46,20 @@ public SolutionInterpreter( Solution solution, boolean appendWon )
         this.command = null;
         this.untilCount = -1;
     }
+    
+    public static SolutionInterpreter getNothingPlaying()
+    {
+        SolutionInterpreter nothingPlaying = new SolutionInterpreter( SolutionParser.parse( "1" ) );
+        nothingPlaying.emptySteps = true;
+        return nothingPlaying;
+    }
```
andybalaam: Better to pass in emptySteps in the constructor and provide a constructor that defaults it. Then it (emptySteps) can be final. Also can we think of a more descriptive name? Continue past end?

andybalaam: If you still want getNothingPlayer I suggest a rename to createNothingPlayer. Get is both a horrible word and usually used to mean access a member variable.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
colonelfazackerley: >  I was expecting these type switch classes to only be implemented inline.

I don't understand. Sorry, you will have to spell it out for me.

GeneralPhysics needs to make calls back to swing, but render is not allowed to depend on swing, so I need an interface in render (or engine) that swing can implement.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
colonelfazackerley: I have not written it yet, but when the solution runs out of steps, and the level has not finished the level needs to change to emptySteps true. So I did not make it final.

continuePastEnd does not sound right for a SolutionInterpreter that starts empty (everything apart from demo mode).

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
andybalaam: Yeah I managed to figure out that you need an interface to reverse that dependency but I was suggesting making onespecifically rather than reusing TimeStepActionTypeSwitch. Those type switch interfaces were intended to be used like switch statements so I was surprised to see one used like s normal interface. That doesn't mean it's wrong but the fact that not all the interface methods are used is maybe a bit suspicious.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
andybalaam: Ah I thought emptySteps meant "send empty steps if you've run out" but it sounds like it means "send empty steps from now on" so my comments are irrelevant.

colonelfazackerley: OK I will create a new interface. How about calling it UiPlayAction?

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
colonelfazackerley: OK. I reused it simply because it vaguely seemed to fit. The purpose of it has not been clear to me.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
colonelfazackerley: Or perhaps UiPlaybackAction, which is more suggestive of demo mode.

colonelfazackerley:
I just noticed the branch name I am using here is wrong "rec..." instead of "play..."

andybalaam:
> clicking while a solution is running can make you run out of tokens, and the solution fails. i think this is correct.

Me too.

andybalaam:
> a solution that runs out of steps throws a null pointer exception. I think the interpreter should just go quiet and let the level run.

Yes.

andybalaam:
> a solution with an excessive number of waits on the end is fine.

Yes, but unsettling that this is different from what happens in unit tests - could lead to confusion when trying to diagnose unit test failures.

andybalaam:
Ooh, that leads me to a thought: could a failed unit test print out a command line you can type to watch the solution play out?  (And do we need an extra `--strict` switch that means fail if you have extra waits?)

andybalaam:
> if the solution has tokens placed illegally on a block square, it is ignored. i think it should probably stop with an exception.

Sounds good.  You could check whether the drop resulted in the number of tokens decreasing, and fail if not?  (This would also fail when you had run out.)

I would favour matching the behaviour of the tests as much as possible in our behaviour here, because a) we will use this to debug failing tests and b) we don't want weird solutions floating around that have extra drops that don't work etc.

rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
andybalaam: Hmm.  I should document it.

rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionInterpreter.java
andybalaam: Sounds good.

andybalaam:
Merged because this is totally cool and ready for merge, but we may want to continue some discussions.

andybalaam:
I have a creeping feeling I would prefer the solution interpretation to happen further "out" i.e. somewhere in or near to the Swing code, rather than right "in" the GeneralPhysics, but to flesh this out I will need to play with the code myself, so I'll shut up about it until I have some coherent thoughts.

PR #286, opened 2016-01-13
debug output once per world step only. print something immutable so e…

### LINKED ISSUE #272: Bug: Rabbit state debug output prints every frame.
Should be every world step.

colonelfazackerley:
…ach rabbit can be tracked

fixes #272


andybalaam:
Thanks and sorry for breaking it.

PR #291, opened 2016-01-16
Speedup

colonelfazackerley:
This works. I just need to 
- make an exception for hyperluminal rabbits
- change the button image as discussed in #1 

I have tried a few values, and I like a 3x speedup. We could make this configurable via  a manual edit of the config file.


andybalaam:
Looks good to me. Are you requesting a merge yet?

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/GameUi.java
```diff
@@ -425,6 +427,17 @@ public void actionPerformed( ActionEvent evt )
         } );
 
         MenuTools.clickOnKey( menu.pause, "pause", KeyEvent.VK_P );
+        
+        menu.speed.addActionListener( new ActionListener() 
+        {
+            @Override
+            public void actionPerformed( ActionEvent evt )
+            {
+                gameLaunch.setSpeed( speeds[speedIndex = -1 * ( speedIndex -1 )] );
```
andybalaam: Mad skillz. I'd prefer the modification of speedIndex be on the line before. Also, if you're going to be that clever I think it deserves a comment.

colonelfazackerley: :)
I am not trying to make it hard to understand. I will split it up a bit.

colonelfazackerley:
dont merge yet. i will add pause and mute icons.

also let me know if i should make the speedup factor configurable.

colonelfazackerley:
so, merge now, or let me know if i should make the speedup configurable.

andybalaam:
Config is bad. We should pick the best value and go with it. I presume we need some work to enable this in Android?

colonelfazackerley:
Yes, the buttons I added are in swing. The change to physics and the exception should apply. I still can't test android, so I did not change that.

PR #301, opened 2016-01-22
Void marks outside the world's edge

### LINKED ISSUE #176: Enhancement: Visible edge of world
The edge of the world should be more visible so that the player knows not to try bridging or climbing into it.

I suggest that a slightly different background is drawn that gives an impression of being in space, or in some sort of vaporiser.

The same graphic could be used for the enhancement of "death squares", so it might be worth making sure the image works ok ad a standalone square.

colonelfazackerley:
closes #176


andybalaam:
This is brilliant work.  I'd prefer not to choose a random marker inside the SpriteAnimator constructor, but pass it in from outside.  I'd also prefer if the same level always had the same surround, somehow.

andybalaam:
2 ways for the same level to have the same surround:

1. Use the same random-number generator we use to choose block variants to choose the surround
2. Specify the void type in the level file

Thoughts?

andybalaam:
Snapshot 0.7.1.1 uploaded to http://artificialworlds.net/rabbit-escape/snapshots/

Issue #303, opened 2016-01-23
On Android void flips between variants each time step


andybalaam:
Since #301 the void around the level appears but it randomly flips between variants each time step. (On Android)

colonelfazackerley:
It did that in swing to start with. then I stored the VoidMarker.Style in SpriteAnimator and it was OK. I guess the Android code must make a new SpriteAnimator each step.

colonelfazackerley:
I will do a pull request to fix. 
- Store the void style in the world. 
- Support parsing void style from rel files.
- if no void style is chosen. do something.... maybe make a number from the world name, and use the remainder after dividing by the number of styles to pick one. then worlds with the same name always get the same edging.

colonelfazackerley:
Oh, is engine allowed to have dependencies on render?

colonelfazackerley:
No, it's not. Should I move the Style enum from VoidMarker to somewhere in engine, possibly World?

andybalaam:
Yes, world seems a sensible enough place.

colonelfazackerley:
I have made another class for it. I can move it to World. I should have the pull request ready this evening.

andybalaam:
Another class is fine too.

colonelfazackerley:
I think this is fixed. Just needs testing on Android.

BTW, could you write a wiki page on compiling for Android. I know you told me once by email or something, but I am struggling to dig it up.

colonelfazackerley:
There is some stuff in INSTALL.md, but some more details (like how to install android studio) would be handy.

colonelfazackerley:
I tested this on android when looking at the speedup button: it does not flip. please close this issue.

andybalaam:
Will make a snapshot when I get a chance - thanks!

andybalaam:
New snapshot 0.7.1.2 up at http://artificialworlds.net/rabbit-escape/snapshots/

PR #306, opened 2016-01-23
store void style in World. generate reproducible styles based on worl…

colonelfazackerley:
…d names

I think this should address #303


colonelfazackerley:
I am just going to tweak so you dont need shouty caps in the rel to declare the style.

colonelfazackerley:
and implement the TODO.

ready to merge, i think.

andybalaam:
Excellent. I really feel like you're starting to code in my style. This may not be a good thing :-). 2 suggestions:
1. You can use name.hashCode or even name.length to get a hash of the name to choose a consistent void,  but if you prefer it how it is I'm ok with it. 
2. I'm pretty sure there's a join function in Utils that can make the list of styles for the exception. 
Happy to merge now or wait for updates.

colonelfazackerley:
I can be taught to perform simple tasks.

Java spec says
>  Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified. This integer need not remain consistent from one execution of an application to another execution of the same application.
 
I would like the same void style after a program restart.

Also, for short names and increasing numbers of styles name.length may not cut it. Lets stick with my homebrew hash.

colonelfazackerley:
There was not quite the right join, but there is now. Good point.

I think this is OK to merge.

>  I really feel like you're starting to code in my style

I am sure I will make your eyes bleed again soon.

andybalaam:
Good point about hashCode. I'd prefer the home made hash to be a separate function. Please can we have a unit test for the new Util (which is great by the way)

PR #307, opened 2016-01-24
@Override in GitHubIssueDialog

colonelfazackerley:
Not sure why Eclipse was not warning me about this before.

BTW, it may be a bit fiddly to get 028b2ebc629478489b4ed00991b0175d959a9a21 working. This should work (assuming that you already have that commit and 'git status' says you have nothing to commit)
- ```rm -r rabbit-escape-ui-text/.settings``` and the same for engine, text and render.
- in eclipse RMB on each project/'settings'/'Properties'/'Java Compiler'/'Settings/Warnings'/'Enable project specific settings'/OK for each project
- close Eclipse
- ```git reset --hard``` to bring back the '.settings' that were rm'd.
- restart Eclipse and you should have the new, strict regime

### DIFF
--- a/rabbit-escape-ui-swing/src/rabbitescape/ui/swing/GitHubIssueDialog.java
+++ b/rabbit-escape-ui-swing/src/rabbitescape/ui/swing/GitHubIssueDialog.java
@@ -65,6 +65,7 @@ public class GitHubIssueDialog extends JDialog implements ChangeListener
         /**
          * @brief
          */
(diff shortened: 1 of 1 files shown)


andybalaam:
Shouldn't the .project file contain that setting?

colonelfazackerley:
It seems that the setting is not in .project

andybalaam:
Stupid eclipse.

PR #308, opened 2016-01-24
Set names

colonelfazackerley:


andybalaam:
I think this will go away when we remember each level you have done based on an unique name but for now it allows ordering without messing up existing configuration so looks good.

PR #309, opened 2016-01-24
Swing framedump for solution playback

colonelfazackerley:
This is slow. On my debian VM its not possible to play and dump frames. Also creating the gif after is slow, even for ImageMagick, so trying to do it in java may not be sensible.

These things may be tractable if the resolution is dropped for the dumped frames.

I think this is useful as it is: it allows the creation of high quality video.

### DIFF
--- a/rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
+++ b/rabbit-escape-render/src/rabbitescape/render/gameloop/GeneralPhysics.java
@@ -49,7 +49,7 @@ public class GeneralPhysics implements Physics
         }
     }
 
-    private static final long max_allowed_skips = 10;
+    private final long max_allowed_skips;
     public static final long simulation_time_step_ms = 70;
 
     public int frame;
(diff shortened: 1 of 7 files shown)


colonelfazackerley:
Here's one I made. It's close to the size limit that github allows.
anim

colonelfazackerley:
A frame that I captured. This level probably isn't any use, unless you fancy something similar for a banner on artificalworlds.
frame_0000098

andybalaam:
I like that frame. We should put it on the home page and play store page.

andybalaam:
I feel sure we can compress them a lot. Maybe convert has options.

colonelfazackerley:
I will work on the banner name level. Try to get a rabbit somewhere doing each activity.

colonelfazackerley:
Just remembered. I need to update the help output.

andybalaam:
Things I'd like to offer: gif bug reports,  tweet the last few seconds (for bragging about victories), visualise unit test failure.

andybalaam:
Animated gif code http://elliot.kroo.net/software/java/GifSequenceWriter/ . Usual license caution applies,  and we would need to check it works with the old version of javs we use.

PR #313, opened 2016-01-26
Metal blocks

GamingInky:
Adds metal blocks as blocks which you can't bash or dig through (in test bashing the bottom test uses a floating bug so if you are going to fix a bug check that test).


andybalaam:
@colonelfazakerley @tttppp comments?

colonelfazackerley:
Is this your first pull request? Congratulations!

Could you put a level or two up to demonstrate the puzzliness of Metal?

colonelfazackerley:
Nice work, btw.

colonelfazackerley:
How about something like this?
metal_block_1
It makes the metal look a bit more different to the non-metal blocks.

I can do something similar to the other variants, if this is approved.

andybalaam:
I like it. @GamingInky ?

rabbit-escape-engine/src/rabbitescape/engine/behaviours/Digging.java
```diff
@@ -51,8 +51,16 @@ public State newState( BehaviourTools t, boolean triggered )
...
+                if ( t.blockBelow().type == Block.Type.metal_flat )
+                {
+                    stepsOfDigging = 0;
+                    return RABBIT_DIGGING_USELESSLY;
+                }
+                else
+                {
                 stepsOfDigging = 2;
                 return RABBIT_DIGGING;
+                }
```
tttppp: Indentation went a bit weird here.

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
```diff
@@ -224,7 +223,14 @@ public boolean isFlat( Block block )
...
     {
-        return ( block != null && block.type == solid_flat );
+        return
+            (
+                block != null && 
+                ( 
+                   block.type == solid_flat 
+                || block.type == metal_flat 
+            ) 
+        );
```
tttppp: Not something to change as part of this pull request, but I think blocks should have a shape, and these methods should be renamed to be about the shape of blocks/interaction between shape of blocks. At the moment creating a metal ramp requires adding metal_up_ramp to a whole load of methods, whereas it should be as simple as setting a block's shape property to ramp, and its material property to metal.

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
```diff
@@ -182,10 +179,11 @@ public boolean isLeftRiseSlope( Block b )
             || b.type == bridge_up_left;
     }
 
-    private boolean isSolid( Block block )
+    public static boolean isSolid( Block block )
```
tttppp: Since this is now public it might be useful to add some Doxygen to explain what the method does. I've played quite a bit of rabbit escape, but I can't say for sure why anyone cares if a block is "solid" or not (and I can't think what qualifies as a non-solid block).

tttppp:
Really nice change! I need to find some time to make a few new levels using brollies and metal!

GamingInky:
I really like the image @andybalaam and @colonelfazackerley.

colonelfazackerley:
I will give all 4 metal block variants the same treatment, and do a pull request into your branch.

GamingInky:
I tried your level #316 @colonelfazackerley and it was harder than it looked.

colonelfazackerley:
@GamingInky 
I thought the metal made it a bit different. What did you think?

GamingInky:
yes it has a intresting solution (could you add the solution in the level file)

GamingInky:
I really like the rainbow metal block

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
andybalaam: Yes, and the more stuff we add the clearer it is that we need this.

rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
andybalaam: Yes. I think it basically means there's something there and it's not a bridge

GamingInky:
I rebased this.

andybalaam:
Excellent work

PR #314, opened 2016-01-26
Dont pull this one!

colonelfazackerley:
just put this up so the android studio project file fighting may be easily visualised.


andybalaam:
If you revert all the idea files to their previous state (with idea closed) then start idea how much gets updated?

colonelfazackerley:
One file is changed. The unfortunate side effect is that the code wont compile anymore. I then closed the project, and re-opened. Some time consuming things happen and then the project works again. Now loads of the files are changed again.

colonelfazackerley:
maybe we can gitignore everything here and then very tentatively bring the odd file back in.

the git pain from this is much worse than the pleasure from shared code formatting and warning settings.

andybalaam:
Yes and code formatting pain can be fixed with command line tools but what I do care about is how easy it is to get started from a clean environment.

andybalaam:
Let's track this in #318

PR #323, opened 2016-01-27
Metal blocks

colonelfazackerley:
@andybalaam  
I was going to do this pull request into @GamingInky's fork, but I branched from master more recently, so it was confusing.


colonelfazackerley:
I have xcf files with layers for these PNG files. Do you want those?

andybalaam:
Yes please maybe in a subdirectory of images-src called "manual" because we won't be auto generating the pngs from them.  (

andybalaam:
Hmm maybe we could...)

colonelfazackerley:
I think xcf files are lost. I was using a linux VM under windows. the VM's virtual disk appears to be borken. (possibly by the partition resizing I did in changing to a dual boot). in my VM, x windows wont start and I can't even get a terminal.

andybalaam:
Oh dear. We can recreate then if needed.

andybalaam:
It is telling me I can merge but I'd feel less nervous if you rebased. Let me know if that is tricky.

PR #327, opened 2016-02-12
Fire

### LINKED ISSUE #300: Enhancement: Discussion: fire
A fire burning in a square would crisp bunnies entering it. It would not have the same complexity as water. Perhaps implement fire first. Of course, if we have fire and water, the water would put out the fire, introducing puzzling opportunities.

colonelfazackerley:
closes #300


andybalaam:
Will take a bit of reviewing but I will get on to it.

colonelfazackerley:
Mostly pictures.

I did sneak a couple of other bits in that should probably have been separate pull requests. Problem is I end up creating merge conflicts between my branches.

andybalaam:
One comment: I'd prefer logical units to come together in a single commit e.g. the unit tests and the real code.  I'm ok with e.g. images going in first because I'm sure they won't break anything and it makes it easier to review but where two pieces of code really belong together (e.g. parsing 'A' and writing out 'A') I'd like to see them together.

andybalaam:
I just tried this out on Galton Box and it is really cool.

colonelfazackerley:
If something is a lot of work, I want to commit and push, to back it up. Is a feature, and then tests for the feature OK as subsequent commits?

andybalaam:
I'd prefer them as a single commit.  You could achieve this using `git rebase -i` then `git push -f` to squash commits together before making the pull request.

On the other hand, I'm ok with separate commits if it's the way it works for you.  The only thing I'd strongly like to avoid is commits that break the build (which I don't think you have done).

andybalaam:
Snapshot 0.8.0.1 uploaded: http://artificialworlds.net/rabbit-escape/snapshots/

PR #333, opened 2016-02-16
update TODO

colonelfazackerley:
should issues be made for the last few things and the TODO file be removed now?


andybalaam:
Yes. I was planning to do that but feel free.

andybalaam:
Thanks again

PR #335, opened 2016-02-18
cheat mode

colonelfazackerley:
- tap out the rhythm from Holst's Mars on the screen, where you drop tokens.
- complete a level
- all levels in that set are now available

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/config/TapTimer.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/config/TapTimer.java
@@ -1,0 +1,46 @@
+package rabbitescape.engine.config;
+
+public class TapTimer
+{
+    /** Tap or click time intervals in ms */
+    private static long intervals[] = new long[10]; // java inits array to zero.
+
+    // Holst's Mars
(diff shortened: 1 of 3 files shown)


colonelfazackerley:
I made a small ogg with the rhythm needed. I can't upload it here...

Perhaps @andybalaam could put it on artificialworlds.net?

andybalaam:
I like it. Remind me if I forget. We should probably have a note about this on the wiki.

andybalaam:
I just finally got around to listening to the music to figure this out (I failed to get it right from the code).  When I tap out the rhythm in Easy 1 it prints Mars to the console, but even when I go out to the main menu and back in to Easy again I can only get into level 1.  Am I doing something wrong?

tttppp:
I really like this idea as something which can be done in Swing and android. We may need a different cheat for the text UI? I was thinking maybe text adventure style:
```
> say "Mars"
```

colonelfazackerley:
You have to complete a level in a set to get the menu to refresh, then the set becomes available.

Good idea, re text mode.

andybalaam:
Documented here https://github.com/andybalaam/rabbit-escape/wiki/Tips-and-Tricks

PR #336, opened 2016-02-18
rabbit ears bounce a bit when walking left

colonelfazackerley:
if this is approved, i will change walking right too. when the head is at the bottom of the motion, the ears bounce a bit.
animation


andybalaam:
This is cool. We might need similar for all the up/down slope stuff too.

PR #337, opened 2016-02-18
Exception msg when solution fails gives command line to see the level…

### LINKED ISSUE #281: When level test fails, print the command line to view what happens
As mentioned in #279, when a test for a level solution fails, print out the exact command line you can type to watch the level play out in the Swing UI.

colonelfazackerley:
… run in swing. closes #281


andybalaam:
Since we know the level and solution can't we calculate this string in a method of the exception?

colonelfazackerley:
I am not sure if the exception knows the set directory name ('01_easy' etc). I can look.

colonelfazackerley:
Did it all in the text file. Need to fix the tests.

colonelfazackerley:
ready. 

BTW: I think testing the exact text of the message for an exception is a step in testing too far.

It makes updates painful. Dont think it adds much. Wont it fail for different languages?

andybalaam:
Perhaps there are too many tests testing the exact text but it does allow us to check that the translation process really works. 

The test cases hard code the language to US English so should pass no matter what the locale of the machine they are running on is.

Issue #318, opened 2016-01-26
Git fighting over Android studio project files


colonelfazackerley:
I suggest using gitignore on the '' rabbit-escape-ui-android/.idea" directory and maybe a few other files. Put the content from there in a zip that can be downloaded from github to get folk started.

colonelfazackerley:
#314

andybalaam:
It would be nice to do better than that. The IDEA folks reckon you can share quite a lot of stuff from that directory. It includes e.g. code formatting and warning settings that weer would like to share. One problem seems to be that it changes when you get a new version of Android studio. If that is the main problem I am happy to check in often and tell people to keep up to date if they don't want to ignore idea stuff.

andybalaam:
If we get some basic instructions done as specified in #319 and #320 then I think we should .gitignore .idea.  Hopefully we can export some code formatting and warning settings and show how to import them during setup.

andybalaam:
I do hate setup steps though, so if we can provide Makefile targets that make things right, that would be better.

colonelfazackerley:
The diff from my latest studio install
```diff
diff --git a/rabbit-escape-ui-android/.idea/codeStyleSettings.xml b/rabbit-escape-ui-android/.idea/codeStyleSettings.xml
index 436aee6..ba48988 100644
--- a/rabbit-escape-ui-android/.idea/codeStyleSettings.xml
+++ b/rabbit-escape-ui-android/.idea/codeStyleSettings.xml
@@ -34,6 +34,34 @@
         <AndroidXmlCodeStyleSettings>
           <option name="USE_CUSTOM_SETTINGS" value="true" />
         </AndroidXmlCodeStyleSettings>
+        <Objective-C-extensions>
+          <option name="GENERATE_INSTANCE_VARIABLES_FOR_PROPERTIES" value="ASK" />
+          <option name="RELEASE_STYLE" value="IVAR" />
+          <option name="TYPE_QUALIFIERS_PLACEMENT" value="BEFORE" />
+          <file>
+            <option name="com.jetbrains.cidr.lang.util.OCDeclarationKind" value="Import" />
+            <option name="com.jetbrains.cidr.lang.util.OCDeclarationKind" value="Macro" />
+            <option name="com.jetbrains.cidr.lang.util.OCDeclarationKind" value="Typedef" />
+            <option name="com.jetbrains.cidr.lang.util.OCDeclarationKind" value="Enum" />
[...]

PR #339, opened 2016-02-19
ignore android studio project files

colonelfazackerley:
i had to 'git rm' some files before this worked. android studio worked OK after this.


andybalaam:
Are you sure ignoring build.gradle is ok? It looks like it won't change to often. ..

colonelfazackerley:
I did rm build.gradle, and it came back fine. if you check the diff in #318, you can see it's just this
```diff
-        classpath 'com.android.tools.build:gradle:1.0.0-rc2'
+        classpath 'com.android.tools.build:gradle:1.5.0'
```
perhaps you can upgrade to 1.5.0, and there will be no changes to the file? then we dont need to gitignore it.

andybalaam:
Let's try that and if it becomes too annoying we will ignore it.

andybalaam:
I think I have addressed this. Reopen if not.

PR #342, opened 2016-02-20
Refactor block type into block material and block shape.

tttppp:
This is to prepare for blocks having a capacity of water they can hold (i.e. determined by shape).

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
@@ -1,6 +1,6 @@
 package rabbitescape.engine;
 
-import static rabbitescape.engine.Block.Type.*;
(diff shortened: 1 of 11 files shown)


andybalaam:
Superb. Metal bridges anyone?

andybalaam:
Why are the enum values SHOUTING?

tttppp:
Enums always SHOUT! Apparently it's because they're constants: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

Maybe bridges should be sponge, or something porous, as I was planning for bridges to let water pass straight through. Probably best to have as a future enhancement though, as separate water regions in a single cell is making my brain ache!

colonelfazackerley:
Wooden blocks for burning...

Good work

colonelfazackerley:
I see, so in languages that use a preprocessor, it would be confusing to use SHOUTING for anything else.

[ducks]

PR #347, opened 2016-02-21
Refactor Direction as an enum.

tttppp:
I plan to add HERE as a "direction" (without an opposite). This also makes reading error messages in tests easier, as UP.toString() is now "UP" rather than "Direction@1234blah".

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/Direction.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/Direction.java
@@ -1,34 +1,18 @@
 package rabbitescape.engine;
 
-
-public class Direction
+public enum Direction
 {
-    private enum D
-    {
-        UP, RIGHT, DOWN, LEFT
-    }
-
-    public static final Direction UP    = new Direction( D.UP );
-    public static final Direction RIGHT = new Direction( D.RIGHT );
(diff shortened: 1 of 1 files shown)


andybalaam:
A few points:
1. This looks obviously better leading me to wonder why I didn't do it before. It may be just me being an idiot, but it may be because it isn't supported on the Android version we are using. 
2. If you need an enum that contains a new direction I suggest creating a new one rather than expanding this one, unless all the existing uses of this one are going

andybalaam:
Sorry, closed accidentally.

andybalaam:
Maybe I didn't realise you could put a function in an enum. I will check it works in Android and merge if it does.

andybalaam:
This works fine on Android API level 8.  I guess I thought you couldn't have methods in enums, or was having some other kind of brain problem.

andybalaam:
Point 2 above still applies I think.  It's not expensive to introduce a new enum if you want a different list of values.

colonelfazackerley:
please could Direction.xOffset (and y) be ```public final``` instead of ```private ``` changeable.

tttppp:
Thanks for the suggestions. My Water branch now has CellularDirection including HERE, and this is kept as is.  I've also changed to use public final constants.

andybalaam:
Awesome

Issue #258, opened 2016-01-06
Merging new levels into old sets (e.g Brolly levels)


andybalaam:
#226 Gave us Brollies, which are brilliant.

Now, we should have levels that use brollies.

Some of these should be merged into existing level sets (possibly making each set contain 25 instead of 20?)  Otherwise new players won't see brollies, which would be criminal.

I guess for now we could make a brolly-based level set, but we will definitely want to spread them around in future.

But, once we start shuffling levels we will mess up our record of their progress.

Currently, progress is tracked via a number for each set of levels, saying which level has been solved (and all the ones before it).

In future (I think) we should store the names (or ids) of levels the player has solved.  If we mix up the order of levels, the UI should illustrate which levels have been solved (and haven't), and allow you to play the first unsolved level in a set (and all levels before it) but not any others.

This requires:

- Unique ids for each level.  We can probably just use their real names, but this would prevent us from e.g. fixing typos in level names.  Perhaps we could use their file name without the leading number?

- An upgrade script that takes current numerical progress and replaces it with the known IDs of the solved levels.

- Change the code that disables levels to use the new ID-based solution.

Really, to make this feature palatable, we would also want:

- #205 Level names in menu items

and

- #257 Tick in menu next to solved levels

colonelfazackerley:
Breaking the immutability of sets makes me nervous. I am not sure it's worth it.

Inserting level sets seems fine to me.

andybalaam:
I just don't think it's sustainable long term. The first 60 levels I ever wrote are not going to be the best introduction to the awesomeness of the game as we keep making it more awesome.

colonelfazackerley:
fair point. I am persuaded.

If we grow the easy set to, say, 25 levels, are we OK with different size level sets?

the way the config stores progress will need a rewrite. I would also like a cheat, so I can test on different platforms/uis without having to play through.

- level set directories prefixed with numbers so they are in the correct order in file browsers
- the progress stored in config should not be broken if we re-order the level sets (to insert an easier one)

andybalaam:
Yes. And +1 for a cheat mode. I was considering odd behaviour in level easy 1 to unlock all levels. Preferably something fun.

andybalaam:
Other easter eggs would also be welcome.

andybalaam:
[...]

PR #355, opened 2016-03-04
Water

tttppp:
Support for water and pipes. Water is produced by pipes, flows around using cellular automation, drowns rabbits and extinguishes fire.


tttppp:
Opening this for comments, suggestions and so people can see if I've missed testing things.  Also so that we can try to finalise the mechanics are right before it's merged to master (in particular how much water drowns a rabbit on a slope and how much extinguishes a fire).

tttppp:
Also, should bridges direct water? I think I've coded it so that it would be easy enough to make bridges block water, but what do we want to happen?

Once there are a few levels using water then it will be hard to change things like this without drastically altering the difficulty of the level.

tttppp:
We should have a test that water flows around a U bend. Maybe also a bulkhead test and a test that it can't get into an enclosed region (eg. through ramps).

andybalaam:
That close button is too easy to hit accidentally

andybalaam:
Very excited about this. Am going to start reading the code as soon as I can

tttppp:
I started making a few levels today. I'll post them separately, but I think having water makes them feel immediately different from existing levels. I found them pretty fun to try and solve too.

rabbit-escape-engine/src/rabbitescape/engine/behaviours/Drowning.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        int yCoordinate = rabbit.y;
+        CellularDirection directionToCheck = UP;
+        if ( rabbit.onSlope )
+        {
+            // The rabbit's head is at the bottom of the cell above.
+            yCoordinate = rabbit.y - 1;
+            directionToCheck = DOWN;
+        }
+        // TODO Find out why the rabbit's y coordinate is allowed to be larger than the size of the world (see solution for easy-12).
```
tttppp: This is kind of TODONE, but I've left this here as a reminder to find out the limits on x and y (i.e. what should this bounds check say?)

andybalaam:
I am working through, reviewing commit by commit.  All looks really good so far.

andybalaam:
Is this just me?  I saw you ignored this test, but has it sneaked back in?

```
    [junit] Testcase: States_must_have_animations_and_frames_must_have_images took 0.27 sec
    [junit] 	Caused an ERROR
    [junit] rabbitescape.render.AnimationLoader.AnimationNotFound {cause=null, name=water_region_empty}
    [junit] rabbitescape.render.AnimationLoader$AnimationNotFound: rabbitescape.render.AnimationLoader.AnimationNotFound {cause=null, name=water_region_empty}
    [junit] 	at rabbitescape.render.AnimationLoader.load(AnimationLoader.java:99)
    [junit] 	at rabbitescape.render.TestAnimations.States_must_have_animations_and_frames_must_have_images(TestAnimations.java:25)
    [junit] 
```

andybalaam:
Or was it a different one you ignored?

tttppp:
Great. I ignored it, then made some primitive animations, then enabled it again. I saw it fail several times locally, but in Eclipse it told me the problem was with fire_a_1, which I could see. This error is with the animation state water_region_empty, which isn't meant to have an animation. I'll have a look how to fix that at some point, or I can ignore the test and it will be fixed with colonelfazackerley's merge.

andybalaam:
We'll need something to make the test pass before I can merge.

andybalaam:
An explicit exception is ok if it should not have an animation.

andybalaam:
Running `make clean test` should have completely predictable behaviour, unlike Eclipse.

tttppp:
I've fixed the failing test now by adding water_region_empty as an explicit exception to the rule that states need animations.

andybalaam:
Cool, thanks. I am still working through the reviews but it all looks beautifully clear so far. Lovely simple diffs.

rabbit-escape-engine/test/rabbitescape/engine/logic/TestDrowning.java
```diff
@@ -0,0 +1,224 @@
...
+            "#nn  " + "\n" +
+            "#n   ",
+
+            "# nP#" + "\n" +
+            "#nnn " + "\n" +
+            "#nn  " + "\n" +
+            "#N   "
+        );
+    }
+
```
andybalaam: This is an epic rabbit tragedy in ASCII-art JUnit test form

andybalaam:
This is awesome.  I have really enjoyed reading the diffs!

andybalaam:
Finally got it running, and even with the placeholder graphics, it looks amazing.  I am really excited about the levels we can make with this.

I've got lots of comments, but this has been hanging around far too long (because it took me so long to review) so I want to merge it and fix the comments as separate pull requests.

andybalaam:
Here are my comments, in arbitrary order:

1. The diffs are really clear and simple, but I would prefer to see the tests for a piece of code in the same commit as the code.  Reasons:
    * pushes towards a test-first approach, which I think makes for better code in most cases.
    * helps me see how well-covered each piece of code is when I'm reviewing
    * each git commit is self-contained and covered by tests

2. Test names should be descriptions of behaviour e.g. `Max_of_two_equal_numbers_is_that_number` or `Rabbits_die_when_they_fall_into_water` as opposed to descriptions of the code they cover e.g. `Max_works`.  This may mean you have more, smaller tests, but I basically think this is a good thing.

3. Keep line length below 80 characters, including comment lines.  If you don't like this, feel free to hate me in silence.

4. In WaterRegion can we keep the members connections and flow final by using `connections.clear()` instead of `connections = new Blah()`?  I love final.

5. Should we display water in the text UI?  I think so, because otherwise water levels are very confusing.

6. Is water state saved and loaded with the World?  This will be needed to play on Android e.g. when you rotate the screen.  I think it probably is but want to check you have considered this.

7. TestTextWorldManip.Round_trip_basic_world() should include a "P" symbol to make sure it gets saved and loaded properly.  This class is also where we should test save and load of water state (point 6).

8. We could do with covering more cases in TestWaterRegion.

9. When a block is added or removed, is it sufficient to recalculate only that point in the water table - should we also recalculate its neighbours?  Probably not.  (But if so, write a failing test for this before fixing it.)

10. Test names should start with capital letters.

Excellent work.

andybalaam:
I made snapshot 0.8.0.7 containing this code: http://artificialworlds.net/rabbit-escape/snapshots/

colonelfazackerley:
Your comments should be on the wiki.

colonelfazackerley:
Many of them are general code style, not specific to this.

andybalaam:
Good point. I'll do that.

andybalaam:
I write something here: https://github.com/andybalaam/rabbit-escape/wiki/Code-style

Please comment and/or fix.

colonelfazackerley:
Looks OK. Nice royal "we".

I have broken that 80 char limit all over the place. I have been using
80: think about a newline
100: hard limit.

tttppp:
Two things which aren't on that page: uncoded characters and line endings. Also it might be worth mentioning Doxygen there, if it isn't anywhere else.

Some of "we" think the style is a little awkward. Particularly the combination of spaces everywhere and 80 chars per line. Anyway, it's not the worst style I've ever seen, and it's a million times better than not having a style at all.

tttppp:
Also while I'm complaining, Eclipse doesn't support ctrl left/right-ing through test names that are in weird Underscore_case.

andybalaam:
80 characters: this project was created to give me relief from the things that annoy me in other projects, so you just have to put up with me being unreasonable. 

Also, it fits on the screen. 

And you can diff it. 

And it  discourages lazy MassivelyConcatenatedNamesObjectFactoryListenerFactoryManagerUtilsSingletonSupervisor

andybalaam:
And you can email it.

colonelfazackerley:
I am not arguing for a higher line length limit. I felt the need to mention that I broke it before it was codified.

andybalaam:
Yeah, np. I have to make the transition from 1-man project to team.

andybalaam:
Underscore separation in test names helps Mark them out as proper sentences. They are only ever mentioned in one place so the need for brevity is not there. The fact that eclipse handles them badly is unfortunate but hey lots of things about eclipse are unfortunate.

andybalaam:
Much of this stuff comes from Kevlin Henney. I'll try to dig out a link sometime.

colonelfazackerley:
https://en.wikipedia.org/wiki/Kevlin_Henney
http://www.curbralan.com/
http://asemantic.blogspot.co.uk/
http://www.two-sdg.demon.co.uk/curbralan/papers/jaoo/ValueBasedProgrammingInJava.pdf

probably some more relevant links...

andybalaam:
@tttppp line endings? I would have thought git would sort those for you?

andybalaam:
On tabs vs. spaces: in a perfect world we could use tabs but we are not in a perfect world and tabs are incredibly painful. They don't email properly, they are different for everyone, and they get messed up easily.

tttppp:
In order of decreasing value of discussion (these are all pretty low value discussion points - all these things have been discussed before):

Line endings: I didn't look for a .gitattributes. If that's there then no need to mention.

Emailing code: This seems fairly specious. Rather than email the open source code, why not use Github? Besides, if you want to email diffs, which is much more sensible than emailing a single file of the code itself, then you need to use a 77 or 78 char limit. On top of this it's probably worth leaving a few chars so email clients can add indentation. Basically email is not a sensible way to share code. Doing anything to encourage email code sharing seems bad to me. Besides, email has attachments for sharing content which is not suitable for the body of an email.

80 char limit: I don't have anything against the number 80, I just think that given some other choices (spaces to pad everywhere, spaces for indentation), it provides poor code readability.

Tabs v spaces: Tabs displaying differently for different people is a huge positive thing. I like to have tabs displaying differently in different clients. I set my tab width to one thing in an IDE and another in a diff client. Even better, my decision doesn't affect anyone else. (Obviously I'm for tabs for indentation, spaces for alignment. Tabs for alignment is a fools errand).

Test naming: Test methods are readily visible due to the annotation. They don't need anything more. Sentences should be in javadoc if they are needed to explain the test. This has the added advantage that you can write more than one sentence to describe the test, it's purpose, or anything else that may be important to future developers.

I'm certain that none of this will convince you to change your mind, and I'd much rather have a fixed style than a mishmash. A defined style is great to avoid edit wars in the commit history, and helps avoid wasted hours spent pondering the best place to insert a line break.

andybalaam:
I totally agree that the existence of a standard is more important than the details. Rest assured you have not convinced me and this discussion would be best continued over a pint and decided via thumb war.

andybalaam:
Argh.  Tried and failed not to respond further :-)

Side note: a lot of the tradition behind my preferences comes from the Linux kernel community, but I have to be careful advocating all of their practices since their tabs/spaces policy is about as wrong as any tabs/spaces policy can be.

> Line endings. I didn't look for a .gitattributes. If that's there then no need to mention.

Actually, on this point it is well worth following up.  I don't understand.  I thought Git would make text files on Windows come out with \r\n and on Linux/Unix/OS X with \n?  I.e. I didn't think there was anything we needed to do.

> Emailing code...why not use Github?

I'd rather not be too dependent on specific tools.  GitHub is proprietary and we all saw what happened to Sourceforge.  Even depending on Git seems risky since some day there's sure to be something better coming along.  Meanwhile, if we wanted to use Git without GitHub, email would probably be vital, and Git has good support for sending patches inline in emails that hook up nicely with scripts that receive and apply patches automatically.  Most of these tools don't work with attachments, since the culture where they grew up sees attachments as unnecessary overhead (two clicks where one would do).

Of course, these are all long-term, and deeply serious concerns that are probably out of place in such a small and insignificant project as Rabbit Escape.

>  then you need to use a 77 or 78 char limit. On top of this it's probably worth leaving a few chars so email clients can add indentation.

Yes, you are probably right.  75 or 77 seems sensible.  But I don't think you're proposing this ;-)

> 80 char limit: I don't have anything against the number 80, I just think that given some other choices (spaces to pad everywhere, spaces for indentation), it provides poor code readability.

You mean because lines get wrapped in the middle of a piece of logic?

I think limiting line length and indentation level improves readability by discouraging over-complex lines, and encouraging separating logical units of work into separate, named entities like functions.  In a Java-specific context it also discourages lazy epithet-concatenated names, hopefully encouraging more thoughtful and concise names.

> Tabs v spaces: Tabs displaying differently for different people is a huge positive thing

Tabs just mess up when you paste them in to different places e.g. web forms/forums/blogs, web page source code, different markup schemes, emails, remote desktops.  In principle, allowing the same code to look like different indentation widths for different people might be a good thing (although I might argue about that too :-) but in practice it just causes pain when you are trying to communicate about the code.

> (Obviously I'm for tabs for indentation, spaces for alignment. Tabs for alignment is a fools errand)

I found the Kevlin Henney article I wanted to link to, but it's private to ACCU members.  This video contains some of the stuff he is talking about: https://vimeo.com/97329157

He argues quite convincingly for "sustainable spacing" i.e. layouts that don't degrade over time when code is refactored or otherwise modified.  This basically leads to the conclusion that alignment in general is a fool's errand.  It took me a while to accept this because I really like lining things up, but I think I am mostly convinced, except in some very special cases (like e.g. lining up numbers so the tens, units etc. are aligned).

Anyway, yes, obviously, where we do alignment we should use spaces.  But are you really advocating for lines that start with tabs for indentation, and then begin using spaces where they begin being about alignment?  Personally, because I always have whitespace visible in my editor and am fairly upset by "wrong" whitespace, I would probably mostly get this right if that were our coding style, but I haven't met many other people who are that committed to "correct" whitespace.  I think most people wouldn't even bother understanding what you were trying to say, and ignore you :-)

Further note: I think it might be hard to enforce this mixed tabs-for-indentation, then spaces for alignment rule via an automated tool?  Since the only thing we agree on is that there should be a standard, automating it seems the best way to avoid spending too much time arguing.

> Test naming: Test methods are readily visible due to the annotation

I'm not trying to make it easy to find tests, I'm trying to promote a different culture when naming tests from naming other methods.  A visibly different style helps to make this differentiation.

I want test names to be full sentences that describe the expected behaviour.  There is [weak evidence](https://whathecode.wordpress.com/2013/02/16/camelcase-vs-underscores-revisited/) that it is easier to read underscore-separated text, but the real advantage for me is that by making it a different style we allow ourselves to adopt a different writing style: in this case allowing very long names up to about 60 characters, which would be horrific (in my opinion) in camel case.

> Sentences should be in javadoc if they are needed to explain the test.

Documentation is always wrong.  The name of the test has the advantage that it is code, so much more likely to be updated by a programmer, and that it is about the right length (1 line) for a description of behaviour at about the right level.  If we allow arbitrary-length descriptions of tests, it becomes easier for us to write tests that test multiple things at the same time.

> I'm certain that none of this will convince you to change your mind

Sorry, yes (except the 77 character thing).

I feel really dumb for wasting so much of your time and my time on this, sorry.

I'm off to work on #258

andybalaam:
@tttppp I'd like to see further discussion and/or pull requests on the non-style parts of https://github.com/andybalaam/rabbit-escape/pull/355#issuecomment-197016624 when you get a chance.

andybalaam:
Just skimming through that video, it contains the motivation for most of my opinions.

PR #365, opened 2016-03-10
Future levels

tttppp:


rabbit-escape-engine/src/rabbitescape/levels/development/paradise/Planel-Paradise.rel
```diff
@@ -0,0 +1,25 @@
+:name=Planet Paradise
+:description=Don't lose your gravity and fall off the planet just get to the core.
```
andybalaam: Typo in this filename

rabbit-escape-engine/src/rabbitescape/levels/development/unsorted/Oh-we-do-like-to-be-besides-the-treeside.rel
```diff
@@ -0,0 +1,26 @@
+:name=Oh we do like to be beside the tree-side
+:description=Wait, can rabbits climb trees?
+:author_name=tttppp
```
andybalaam: Extra s in filename

andybalaam:
As mentioned in the issue, I'd much prefer it if we could load levels from the class path like we do with real levels. This would need an ls.txt file for dir generated by the make file. If you prefer we can merge this and change that afterwards.

tttppp:
As mentioned in the issue I think this does use the classpath, in the same way that the real levels do.  I'll fix the typos now.

tttppp:
Fixed the typos and squashed the commits.

andybalaam:
Yes, I was completely wrong. This is perfect.

andybalaam:
If we want cheat mode to unlock these levels we will need them to be part of the normal menu definition, but this is good for now.

PR #372, opened 2016-03-26
polygons for water in swing and android

colonelfazackerley:
ready.


colonelfazackerley:
I should do android too

colonelfazackerley:
I added code for android, which can be seen in the diff here. I also changed easy 1 to aquifer, and started debugging through android studio. it crashes when I start the level with water in (but is OK for other levels).

The "5: Debug" frame in studio does not give me a stack trace, or say anything useful. I dont know how to find the problem.

colonelfazackerley:
ah, I found the stack trace in monitor.
```
                 Caused by: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
                                                                                       at rabbitescape.ui.android.Dialogs$IntroDialogs.hint(Dialogs.java:95)
                                                                                       at rabbitescape.ui.android.Dialogs$IntroDialogs.show(Dialogs.java:63)
                                                                                       at rabbitescape.ui.android.Dialogs.intro(Dialogs.java:168)
                                                                                       at rabbitescape.ui.android.AndroidGameActivity.onCreate(AndroidGameActivity.java:97)
                                                                                       at android.app.Activity.performCreate(Activity.java:6251)
```

so, it needs a hint, i think. android should be more graceful with missing hints, or tests should insist on hints.

colonelfazackerley:
It runs, but I cant see the water.

It feels like I am close.

colonelfazackerley:
I added some missing stuff from android. does not draw anything. :-(

colonelfazackerley:
I figured out how to use logcat, and that the poly drawing function is not being called.

colonelfazackerley:
There are  no polygons...

colonelfazackerley:
aha. WaterDynamics has a null world. Probably because I make the objects in the wrong order.

andybalaam:
I'm going to need to concentrate properly before I can get involved and help with this. I feel I need to finish off #258 before I can do that. There's no hurry to make water release-ready so we may want to disable it for 0.9 if we feel it needs more bedding in. 

I'm also happy to merge the non Android polygon code if it helps.

colonelfazackerley:
I think I am almost there. but then I thought that before.

colonelfazackerley:
WaterDynamics.step now has a valid reference to World, but still no polygons. Boo.

Gonna play Wesnoth now....

colonelfazackerley:
screenshot_20160328-080317

Yay!

Android was using the WaterDynamics constructor only meant for tests. The other WaterDynamics constructor was also being called, which is why it took me a while to notice.

some tidying to do now

colonelfazackerley:
rebased too

colonelfazackerley:
I think this is ready.

GamingInky:
I cloned water4 and it still used the images

tttppp:
Maybe try a `git fetch` before checking out Water4?

colonelfazackerley:
@GamingInky
I just checked the diff here. Look at spriteanimator. I removed the bit where tttppp's added the sprites.

Not sure why it's not working for you.

GamingInky:
```
$ git clone https://github.com/colonelfazackerley/rabbit-escape.git
<snip>
$ cd rabbit-escape/
$ git status
On branch master
<snip>
$ git branch
* master
$ git fetch origin Water4
 * branch            Water4     -> FETCH_HEAD
$ git checkout Water4
Branch Water4 set up to track remote branch Water4 from origin.
Switched to a new branch 'Water4'
$ make
<snip>
. Compiling
    [javac] /home/kids/Desktop/water/rabbit-escape/rabbit-escape-render/src/rabbitescape/render/Vector2D.java:5: warning: [overrides] Class Vector2D overrides equals, but neither it nor any superclass overrides hashCode method
    [javac] public class Vector2D
    [javac]        ^
    [javac] error: warnings found and -Werror specified
    [javac] 1 error
    [javac] 1 warning

BUILD FAILED
/home/kids/Desktop/water/rabbit-escape/build.xml:40: The following error occurred while executing this line:
/home/kids/Desktop/water/rabbit-escape/build.xml:13: Compile failed; see the compiler error output for details.

Total time: 1 second
make: *** [compile-noui] Error 1
kids@ubuntu-desktop:~/Desktop/water/rabbit-escape$ git status
On branch Water4
Your branch is up-to-date with 'origin/Water4'.

nothing to commit, working directory clean
```

GamingInky:
```
$ java -version
java version "1.8.0_74"
Java(TM) SE Runtime Environment (build 1.8.0_74-b02)
Java HotSpot(TM) 64-Bit Server VM (build 25.74-b02, mixed mode)
```

GamingInky:
@andybalaam showed me how to turn off warnings-as-errors and now it works!

colonelfazackerley:
Looks like my fault. Eclipse compiles without complaining about that.

I could swear I had done ```make clean slowtest```. It will be this evening before I can fix that.

GamingInky:
I looked at the water and it looks amazing and a bit jercky but we can live with that.

colonelfazackerley:
I did a ```make clean``` and I don't get the hashcode complaint about Vector2D. I did ```git diff origin/Water4``` to check if I had not pushed a commit, but that returned nothing.

I don't know why you get it and I don't.

Have to wait for @andybalaam to explain it.

andybalaam:
Could it be java version?

colonelfazackerley:
OK, it was java version. I have purposefully left my *javac* on 1.7 to avoid problems, but left *java* on 1.8 becuase of the superior transparent polygon performance. Is it wise to change javac to 1.8? I am doing this
```
cf@kl:~/re/rabbit-escape$ sudo update-alternatives --config javac
There are 2 choices for the alternative javac (providing /usr/bin/javac).

  Selection    Path                                         Priority   Status
------------------------------------------------------------
  0            /usr/lib/jvm/java-8-openjdk-amd64/bin/javac   1081      auto mode
* 1            /usr/lib/jvm/java-7-openjdk-amd64/bin/javac   1071      manual mode
  2            /usr/lib/jvm/java-8-openjdk-amd64/bin/javac   1081      manual mode

Press <enter> to keep the current choice[*], or type selection number: 1
```

```
cf@kl:~/re/rabbit-escape$ sudo update-alternatives --config java
There are 3 choices for the alternative java (providing /usr/bin/java).

  Selection    Path                                            Priority   Status
------------------------------------------------------------
  0            /opt/java/64/jre1.8.0_77/bin/java                2000      auto mode
* 1            /opt/java/64/jre1.8.0_77/bin/java                2000      manual mode
  2            /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java   1071      manual mode
  3            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1081      manual mode

Press <enter> to keep the current choice[*], or type selection number: 1
```

andybalaam:
Sigh.  Compiling with 1.7 is probably a good idea for compatibility, but 1.8 gives better warnings.  If/when I get around to setting up continuous integration we should probably use both.

rabbit-escape-engine/src/rabbitescape/engine/util/CellDebugPrint.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private static String formatString( String s )
+    {
+        if ( maxLength == 0 )
+        {
+            return "";
+        }
+        return String.format( "%" + maxLength + "s", s );
+    }
+}
```
andybalaam: I'd like to see unit tests for CellDebugPrint, just so I can understand what it does.

rabbit-escape-engine/src/rabbitescape/engine/util/LookupTable2D.java
```diff
@@ -149,4 +149,9 @@ private int i( int c )
     {
         return list;
     }
+
+    public int itemCount()
+    {
+        return list.size();
+    }
 }
```
andybalaam: Any reason this (itemCount) is not called size?

rabbit-escape-engine/src/rabbitescape/engine/util/CellDebugPrint.java
```diff
@@ -0,0 +1,110 @@
...
+{
+    private final Vector<Vector<Vector<String>>> strings;
+
+    private static int maxLength = 0, maxX = 0, maxY = 0, maxI = 0;
+    private static boolean off = false;
+    
+    public CellDebugPrint()
+    {
+        strings = new Vector<Vector<Vector<String>>>();;
+    }
```
andybalaam: Why Vector and not ArrayList?

rabbit-escape-render/src/rabbitescape/render/Polygon.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class Polygon
+{
+    public Vector<Integer> x, y;
+
+    public Polygon()
+    {
+        x = new Vector<Integer>();
+        y = new Vector<Integer>();
+    }
```
andybalaam: Here too: why Vector and not ArrayList?

rabbit-escape-render/src/rabbitescape/render/Polygon.java
```diff
@@ -0,0 +1,45 @@
+package rabbitescape.render;
+
+import java.util.Vector;
+
+import rabbitescape.engine.util.Util;
+
+public class Polygon
+{
+    public Vector<Integer> x, y;
```
andybalaam: Maybe an ArrayList<Point> would be clearer?

andybalaam: Making `x` and `y` public here makes it easy for an outsider to violate an invariant of this class, which is that x and y should have equal length (right?)

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public Vector2D divide( int operand )
+    {
+        return new Vector2D( this.x / operand, this.y / operand );
+    }
+
+    public static Vector2D cellularDirection( CellularDirection d )
+    {
+        return new Vector2D( d.xOffset, d.yOffset );
+    }
```
andybalaam: Should this be called `translate`?

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
```diff
@@ -0,0 +1,99 @@
...
+        if ( o instanceof Vector2D )
+        {
+            Vector2D v = (Vector2D)o;
+            return v.y == this.y && v.x == this.x;
+        }
+        else
+        {
+            return false;
+        }
+    }
```
andybalaam: As Java 8 points out, when you implement equals you are duty bound to implement hashCode.  Something like `31 * x + y` should suffice.

rabbit-escape-render/src/rabbitescape/render/SpriteAnimator.java
```diff
@@ -57,7 +56,7 @@ public SpriteAnimator(
         this.animationCache = animationCache;
     }
 
-    public List<Sprite> getSprites( int frameNum )
+    public List<Sprite> getSprites( int frameNum, WaterDynamics waterDynamics )
```
andybalaam: Why is `waterDynamics` added as a param here?

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
```diff
@@ -0,0 +1,99 @@
+package rabbitescape.render;
+
+import rabbitescape.engine.CellularDirection;
+
+public class Vector2D
+{
```
andybalaam: Rabbit Escape contains 3 classes called `Point` already, 2 of which are identical :-)

However, I think `Vector2D` is a legitimate addition here since the methods on it don't make sense for those other classes.  As a side note, at least 2 of the `Point` classes should be combined.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -0,0 +1,364 @@
...
+        this.region = region;
+        this.world = world;
+        this.dynamics = dynamics;
+    }
+
+    public void setWaterRegion( WaterRegion region )
+    {
+        this.region = region;
+    }
+
```
andybalaam: Do we need `region` to be changeable?  I don't see you changing it anywhere?

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -0,0 +1,364 @@
...
+    public WaterRegionRenderer adjacentRenderer( CellularDirection d )
+    {
+        return dynamics.lookupRenderer.getItemAt( region.x + d.xOffset, region.y + d.yOffset );
+    }
+
+    public boolean adjacentNull( CellularDirection d )
```
andybalaam: `isAdjacentNull` ?

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -0,0 +1,364 @@
...
+        if ( lastHeight - maxHeightChange > targetWaterHeight )
+        {
+            height = lastHeight - maxHeightChange;
+            lastHeight = height;
+            return;
+        }
+        height = targetWaterHeight;
+        lastHeight = height;
+    }
+
```
andybalaam: Is this equivalent?

```
public static int clamp( int val, int min, int max )
{
    return Math.max( min, Math.min( max, val ) );
}

public void updateWaterHeight()
{
    height = clamp(
        targetWaterHeight,
        lastHeight - maxHeightChange,
        lastHeight + maxHeightChange
    );
    lastHeight = height;
}
```

andybalaam: By the way, your impl and mine above both end up with height == lastHeight at the end of the method.  Is that intended?

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -0,0 +1,364 @@
...
+                retY.add( y + 32 );
+                return;
+            default:
+                return;
+            }
+        }
+        Block block = world.getBlockAt( region.x, region.y );
+        int boundaryHeight = calcBoundaryHeight( d );
+        //boolean eitherFalling = isFalling() || adjacentWaterIsFalling( d );
+        //boundaryHeight = eitherFalling ? 0 : boundaryHeight; // If either is falling, then go with zero
```
andybalaam: Can remove comments unless they are useful?

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -0,0 +1,364 @@
...
+        if ( BehaviourTools.s_isFlat( b ) )
+        { // Flat block cells have zero capacity: always full.
+            return true;
+        }
+        WaterRegionRenderer wrr = dynamics.lookupRenderer.getItemAt( x, y );
+        if ( null == wrr )
+        { // Not full of flat block, but no region yet, empty
+            return false;
+        }
+        return 32 == wrr.height && 32 == wrr.targetWaterHeight;
```
andybalaam: I may be misunderstanding, but does this mean blocks containing slopes can't be full?  When might `wrr.targetWaterHeight` not be 32?

rabbit-escape-render/src/rabbitescape/render/gameloop/WaterDynamics.java
```diff
@@ -0,0 +1,157 @@
...
+        {
+            Polygon p = new Polygon();
+            WaterRegionRenderer start = null;
+            for ( int x = 0; x < worldSize.width; x++ )
+            {
+                WaterRegionRenderer wrr = lookupRenderer.getItemAt( x, y );
+                if ( null == wrr || wrr.adjacentWaterIsFalling( HERE ) )
+                {
+                    continue;
+                }
```
andybalaam: Can we make a new `p` here instead of in 2 places?

rabbit-escape-render/test/rabbitescape/render/TestVector2D.java
```diff
@@ -0,0 +1,125 @@
...
+import rabbitescape.engine.CellularDirection;
+
+public class TestVector2D
+{
+    @Test
+    public void Vectors_add()
+    {
+        Vector2D a = new Vector2D( 1, 2 );
+        Vector2D b = new Vector2D( 3, 4 );
+        assertThat( a.add( b ), equalToVector2D( new Vector2D( 4, 6 ) ) );
```
andybalaam: I don't think you need `equalToVector2D`.  I think if you use `equalTo` it will just work.

rabbit-escape-render/src/rabbitescape/render/gameloop/WaterDynamics.java
```diff
@@ -0,0 +1,157 @@
...
+import rabbitescape.engine.util.CellDebugPrint;
+import rabbitescape.engine.util.Dimension;
+import rabbitescape.engine.util.LookupTable2D;
+import rabbitescape.render.Polygon;
+import rabbitescape.render.WaterRegionRenderer;
+
+/**
+ * Manages animation of polygons for water.
+ */
+public class WaterDynamics
```
andybalaam: Maybe this should be called `WaterAnimation`?  I started thinking it controlled the movement of water, not how it looks.

rabbit-escape-ui-android/app/src/main/java/rabbitescape/ui/android/AndroidCanvas.java
```diff
@@ -40,6 +43,24 @@ public void drawLine(
         float startX, float startY, float stopX, float stopY, AndroidPaint paint
     )
     {
-        this.canvas.drawLine( startX, startY, stopX, stopY, paint.paint );
+        this.canvas.drawLine(startX, startY, stopX, stopY, paint.paint);
+    }
+
+    @Override
+    public void drawFilledPoly( int[] xs, int[] ys, AndroidPaint paint )
```
andybalaam: These `androidlike` classes are supposed to be a wafer thin wrapper over the Android APIs, so unless there's a really good reason for it, I'd prefer if you added a `drawPath` method here, and made an `androidlike.Path` interface.  I don't particularly like this interface, but I am keen that these classes contain no complexity at all.  We don't want to make a new drawing API here - we want to simulate the Android one for Swing, and wrap the Android one for Android untouched.

rabbit-escape-ui-swing/src/rabbitescape/ui/swing/SwingGraphics.java
```diff
@@ -81,6 +92,17 @@ void draw( Graphics2D g )
 
             soundPlayer.play( sprites );
         }
+
+        void drawPolygons( Vector<Polygon> polygons, SwingCanvas swingCanvas )
```
andybalaam: This is almost repeated across Swing and Android - can we do it in renderer?

andybalaam:
Sorry I've made so many comments.  Get back to me with feedback about them.

I saw it actually running today and it looks amazing, so I am excited to merge it.

I am worried that the water feature (ho, ho) is going to need delaying since the saving and loading of water state within a level is not done (if I understand right) and this is necessary for Android.  I thought there was an issue for this but I can't find it...

andybalaam:
I think if we do have problems with water, we can just delay shipping water levels, rather than have to back out or disable any code, right?

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
tttppp: I added a constrain method in MathUtils that could be used rather than clamp. Also I added max and min methods for some unknown reason.

rabbit-escape-engine/src/rabbitescape/engine/util/CellDebugPrint.java
colonelfazackerley: Done. 

I will work through the rest of your comments.

rabbit-escape-engine/src/rabbitescape/engine/util/LookupTable2D.java
colonelfazackerley: I like size because it is short. I like itemCount as it is  a better description. size could be the 2D dimension. If you prefer size, we can go with that.

done.

rabbit-escape-engine/src/rabbitescape/engine/util/CellDebugPrint.java
colonelfazackerley: old habit.

I changed to ArrayList.

rabbit-escape-render/src/rabbitescape/render/Polygon.java
colonelfazackerley: changed

rabbit-escape-render/src/rabbitescape/render/Polygon.java
colonelfazackerley: I used 2 arrays because 2 are needed in the end for swing's fillPolygon. However, we have to loop through to scale and offset the coordinates anyway, so we may as well use one list. I agree one list should be neater.

colonelfazackerley: I want to change to float from int too. I had some half working code that smoothed the water surface, and that would look better with float.

tttppp:
Issue for water persisting is #374.

rabbit-escape-render/src/rabbitescape/render/Polygon.java
colonelfazackerley: i refactored. polygon building is in a separate class so the coordinates can be final.

its in a list of Vertex (which is the only float coordinate class)

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
colonelfazackerley: done, though the coordinats are in nominal pixel units and may span the world, not just one cell. so  i used 32 * 10000 * x + y

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
colonelfazackerley: I prefer that name too. Done.

rabbit-escape-render/src/rabbitescape/render/SpriteAnimator.java
colonelfazackerley: Fossil code accident. 

i am now thinking droplets for falling water could just be transparent circles, so this wont be needed. (it might have been if they were sprites).

removed.

rabbit-escape-render/src/rabbitescape/render/Vector2D.java
colonelfazackerley: I agree. I thought Vector2D was a bit different, so OK.

I made an issue about the other duplication.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
colonelfazackerley: WaterRegion instances are re-created every game step. WaterRegionRenderer needs persistent state so water levels rise and fall smoothly (and in the future so water droplets move smoothly).

WaterDynamics.step() should use this to connect WaterRegionRenderer with a new WaterRegion. you are correct that it does not. I need to think about this one.

colonelfazackerley: no, that is all wrong. setWaterRegion can just go. 

done

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
colonelfazackerley: I changed it to adjacentIsNull

if ( adjacentIsNull( ...

flows better.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
colonelfazackerley: much simplified with MathUtils.constrain.

Thanks.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
colonelfazackerley: the height calc takes account of slope blocks. so they are full when at 32, like empty cells.

targetWaterHeight tracks the contents of the WaterRegion from engine. So if a non-block cell is half full (content 512), the targetWaterHeight will be 16. the height lags behind a little to make the height changes look smoother.

rabbit-escape-render/src/rabbitescape/render/gameloop/WaterDynamics.java
colonelfazackerley: i noticed those wrr.drawn* booleans are vestigial, and removed them.

i would not be surprised if this method could be simplified, but I cant see it at the moment.

rabbit-escape-render/test/rabbitescape/render/TestVector2D.java
colonelfazackerley: you are right.

I spent a long time getting this wrong.

rabbit-escape-render/src/rabbitescape/render/gameloop/WaterDynamics.java
colonelfazackerley: I agree WaterAnimation is a better name. done

GeneralPhysics does not make much sense, either.

rabbit-escape-ui-android/app/src/main/java/rabbitescape/ui/android/AndroidCanvas.java
colonelfazackerley: the android Path class is huge and powerful. It does not seem reasonable to attempt to duplicate it in swing when we dont need it. calling it addPath would give the wrong impression. All we want to do is a small bit of what Path can do: polygons.

colonelfazackerley:
I have implemented most of your suggestions, and occasionally been argumentative. I think this is ready.

Do you want these commits fixup'd together, or shall we keep the review history?

i agree this code can go in and cause no problems on droids, if we have no water levels.

rabbit-escape-render/test/rabbitescape/render/TestVector2D.java
andybalaam: :-) always time well spent

andybalaam:
I am working through your responses - all good so far. I'm undecided about squashing the commits. Thoughts?

colonelfazackerley:
I don't mind. If I ran a project it would probably have an extremely messy commit history, until I had an ugly experience to make me reconsider.

rabbit-escape-ui-android/app/src/main/java/rabbitescape/ui/android/AndroidCanvas.java
andybalaam: I would still prefer to implement a small part of the Android interface than make our own 3rd interface that is unlike either the swing or Android one. It makes it easier to understand where the logic is, and hopefully makes it easier to share code between the 2 implementations.

colonelfazackerley: so I should 
- make a Path interface in render with moveTo() lineTo() and close(), ignoring the 30 or so other methods in android.graphics.Path.
- make a drawPath method for swing.

andybalaam: If you can stand to, yes. If you object let's settle it using thumb war.

colonelfazackerley:
> so I should
>
> - make a Path interface in render with moveTo() lineTo() and close(), ignoring the 30 or so other methods in android.graphics.Path.
> - make a drawPath method for swing.

I will get back to this. After the above, I think it will be ready. I am waiting for #401 first, as I know they will merge conflict each other.

colonelfazackerley:
That's it, I am on the drawPath now.

[rimshot]

colonelfazackerley:
I have made it more android like. It is ready.

Let me know if you want it fixup'd together, or if you want the code review left in the history.

andybalaam:
So with this I think water is ready to be featured in future releases, right?

colonelfazackerley:
i think some representation of falling water might be considered essential. There are some glitches too.

this level shows them up. #382

PR #385, opened 2016-04-01
Higher drowning sound

andybalaam:
@colonelfazackerley what do you think of this drowning sound?  The one you made is fairly recognisably a human voice, and lower-pitched than the other voice sound effects.  I don't particularly like what I've done with it (increased its pitch and repeated it over itself), but present it as a possible idea-inspirer?


colonelfazackerley:
yours is better than mine. 

I might try for an even better one. I think yours is a little heavy with the effects.

Also, after considering the pitch issue, perhaps it should be even higher. This would fit with the other effects, like the exit sound.

colonelfazackerley:
I think #394 is better. Whadayareckon?

colonelfazackerley:
#394 was accepted, so I think this can be closed.

PR #391, opened 2016-04-05
insist array by key elements are in order and have no gaps

colonelfazackerley:


andybalaam:
Would be nice to have a test for the out of order case?

colonelfazackerley:
That's not a question. It's like listening to an unnecessary pitch rise at the end of a sentence.

(Warning: using computer under the influence of gin).

andybalaam:
It's a fair cop.

PR #395, opened 2016-04-08
unify coord usage to util.Position. closes #386

### LINKED ISSUE #386: Ugliness: coordinates everywhere
rabbitescape.engine.textworld.Chars.Point
rabbitescape.engine.textworld.LineProcessor.Point
rabbitescape.engine.WorldChanges.Point
rabbitescape.engine.util.Position

colonelfazackerley:


rabbit-escape-engine/src/rabbitescape/engine/util/Position.java
```diff
@@ -15,4 +15,36 @@ public Position plus( Position p )
...
+    @Override
+    public int compareTo( Position other )
+    {
+        if ( y < other.y )
+        {
+            return -1;
+        }
+        else if( y > other.y )
+        {
+            return 1;
```
tttppp: `y - other.y` would cover both not equal cases.

rabbit-escape-engine/src/rabbitescape/engine/util/Position.java
```diff
@@ -15,4 +15,36 @@ public Position plus( Position p )
...
+            {
+                return -1;
+            }
+            else if( x > other.x )
+            {
+                return 1;
+            }
+            else
+            {
+                return 0;
```
tttppp: `x-other.x` would cover all three cases.

tttppp:
Added a couple of comments on some code I only just noticed wasn't written in this change. Woot. Change looks good to me!

colonelfazackerley:
It occurred to me too, but I left it alone.

colonelfazackerley:
I pushed a commit that simplifies compareTo().

colonelfazackerley:
I read the API spec for comparable, and added some more tests, and an equals for Position.

rabbit-escape-engine/src/rabbitescape/engine/util/Position.java
```diff
@@ -15,4 +15,43 @@ public Position plus( Position p )
...
+        {
+            return false;
+        }
+        Position p = (Position)o;
+        return this.x == p.x && this.y == p.y;
+    }
+
+    /**
+     * This may break if we have levels with a dimension greater than 1000
+     * and the class is used for nominal pixels relative to the world origin.
```
tttppp: It won't actually break as such. It's just optimised for x coordinates between 0 and 999. Anyway, good idea to add hash code and equals!

colonelfazackerley:
I see. Hash code collisions do not constitute an error. I will amend that comment.

colonelfazackerley:
done

colonelfazackerley:
i added equals, as the API spec for comparable suggests that compareto should return zero when equals returns true.

andybalaam:
I like compareTo particularly.

PR #397, opened 2016-04-10
use array when parsing string array by key. fixes #389

### LINKED ISSUE #389: Bug: on android 3rd hint is never shown.
I was trying to make it more tolerant of no hints, and my code dropped the 3rd hint. After getting no where for a while I reverted Dialogs.java and found it drops the 3rd hint now, and it's not even my fault.

I will fix this in my Water4 branch.

colonelfazackerley:


tttppp:
The logic in this doesn't look equivalent to what I assumed the old code was meant to do. Are we allowed to just specify hints 1 and 3, omitting hint 2?

Also there's a typo that I'll mark on.

rabbit-escape-engine/src/rabbitescape/engine/textworld/LineProcessor.java
```diff
@@ -155,18 +155,17 @@ public String metaString( String key, String def )
         }
         else
         {
-            ArrayList<String> ret = new ArrayList<String>( temp.size() );
+            String[] ret = new String[temp.size()];
+            for ( int i = 0; i < temp.size() ; i++ )
+            {
+                ret[0] = "" ;
```
tttppp: 0 should presumably be i.

colonelfazackerley: It should. the fact it works with that error makes me wonder if setting the strings to "" can be skipped.

andybalaam: There will be `null` in any entry not initialised and not filled in from things in temp. If we are checking that temp only contains nonrepeating contiguous numbers starting at 1 then I think we are ok, but I'm hoping we can think of code that can make it clearer we're not making a mistake. How about looping an int starting at 1 and increasing by 1 each time and looking up in temp using that int as a key?

andybalaam:
Sorry I didn't spot this in the original review btw.

colonelfazackerley:
@tttppp a pull request of mine was recently merged preventing out of order and missing hints being parsed.

i updated it, using @andybalaam's suggestion.

re: apologies. Mammalian brains are not suited to rigorous checking. It is to be expected.

FWIW: i also apologise for ```ret[0]```. That was pretty bad.

colonelfazackerley:
rebased to remove a note i had already actioned.

tested this using ADB on the bus :)

andybalaam:
I presume we won't see this error message in the wild? Nice fix, thanks.

colonelfazackerley:
> I presume we won't see this error message in the wild? Nice fix, thanks. 

the RuntimeException should never be seen. I cant imagine how it would happen. However, I have had things like that crop up: my imagination is obviously limited. if it does happen, it should be faster to figure out why.

I tested on linux and android.

PR #404, opened 2016-04-21
New Smallworld level: Uno.

tttppp:
This is 7x9, so I think it qualifies for Smallworld.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/levels/development/smallworld/Uno.rel
+++ b/rabbit-escape-engine/src/rabbitescape/levels/development/smallworld/Uno.rel
@@ -1,0 +1,27 @@
+:name=Uno
+:description=This time you only have one of each token.
+:author_name=tttppp
+:author_url=https://github.com/tttppp
+:hint.1.code=(Deqg]\,9q9,Eh\eg#C99n,A6,Dhqh{,?gh,,{#f977Scg,#
+:hint.2.code=-I6-#-T#<V)T-7IIV(#}$TW)B6-Ig~t~&(f(-Tt~-f$-&C["I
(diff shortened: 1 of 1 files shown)


tttppp:
When I thought about this during he night, I realised it's much easier than I intended. Maybe the bomb token should be moved to the bridge token, and the bridge token should be removed. Also the initial delay could be increased a bit.

colonelfazackerley:
I have not tried this yet. Not all levels have to be hard...

tttppp:
I improved the level. It has the same "one of each token" theme, but it's actually possible to go wrong now. I also managed to shrink the level in the process.

colonelfazackerley:
After 10 min, I got 2 out. Yet to hit the target.

andybalaam:
Good work focusing on small world. We need to complete a level set to have something worth releasing, and we are badly overdue for a release.

colonelfazackerley:
We have a lot of new features to showcase. As soon as the overlay and water polygons are sorted, I will focus on levels and sets.

andybalaam:
Woot. We need to test very thoroughly. I still need to figure out how I screwed up the configuration upgrade too :-(

tttppp:
I think I should have put this in staging really, as it's got hints and a solution. Nevermind, someone can fix when they next add a small world level.

colonelfazackerley:
i got 4 out

W=kSWk=~(USjk*}P]!WE(o(:(S~Oh~hSk@E#~~s}*f~}Eh](=~!jWh""S"hW["E~

i think 5 may be possible. i am not going to look at the official solutions quite yet.

colonelfazackerley:
i retract that. i cant see how to get 5 out. i checked the official solution, and it is similar to mine.

andybalaam:
Currently at https://github.com/andybalaam/rabbit-escape/blob/master/rabbit-escape-engine/src/rabbitescape/levels/staging/smallworld/Uno.rel

PR #409, opened 2016-04-25
New level Sloping Off.

tttppp:
Also move Uno to staging.


tttppp:
An easy one.

colonelfazackerley:
I liked it it has novel puzzle elements.

"L}i}RP3!RPldmR}il= 3^}^}RP|mg} !^}&!g=m!Rl l}=i^L3^*i})}m"<}"g},xi}

I think it needs an explode token.

colonelfazackerley:
My solution above is different to yours. Mine relies on the fact that tokens dropped first are picked up first.

tttppp:
It was inspired by the conversations about rabbit ordering that we had. The level would be much easier if the right hand rabbits picked up tokens before the left.

tttppp:
I updated it to make it marginally harder by adding a couple more rabbits and another floater token. Your solution still works, but is slightly trickier! :-)  I also added an explode token, because that can be a satisfying way to get rid of the last rabbit.

tttppp:
I can't see any way to save all ten, but that would make a nice Deja Vu level if it's possible.

PR #412, opened 2016-04-27
update gradle version to 2.1.0

colonelfazackerley:
i do wonder if just ignoring build.gradle would be a good idea.

studio has 2.1.0 on stable, and this comes with it.

### DIFF
--- a/rabbit-escape-ui-android/build.gradle
+++ b/rabbit-escape-ui-android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


colonelfazackerley:
I struggle to not commit this again along with something else. Makes the droid stuff more bothersome than it needs to be.

If you could consider this, that would be great. It's a one liner.

andybalaam:
Does it rebuild if you delete it? In particular is there a command line route to get it back?

colonelfazackerley:
Does the android code rebuild? delete what?

I dont understand the question.

andybalaam:
If you delete build.gradle does it get recreated automatically?

PR #413, opened 2016-04-28
New level: Whack-a-rat.

tttppp:
It's pretty easy, but it's also the thinnest level we're going to get.


colonelfazackerley:
It's a funny idea. I think the target should be 6. The most I can do is 7.

colonelfazackerley:
I got 10 in text mode!

.... and 9 in swing.

tttppp:
I don't think 10 in swing is too hard - particularly compared with some of the monsters we've created! The first rabbit is tricky, but the rest aren't too bad.

tttppp:
I just realised that I missed an opportunity for an enforced cheat mode enabler here. If the rabbits appeared to the right rhythm I think anyone solving the level would automatically open cheat mode. (Is this right?)

andybalaam:
Nice.

PR #418, opened 2016-05-05
cheat mode enables dev levels on swing and droid.

### LINKED ISSUE #366: Cheat mode should enable development levels
Somehow, we should have a mode where you can play the levels that have been committed to git but not yet exposed in the menus.

I propose that when you enter cheat mode, new menu items are made for all the extra directories inside the levels directory appear.

colonelfazackerley:
closes #366 

(i wish i could say "implements" instead of "closes").

ready.

levels need to be labelled by name, but that can be a separate issue.


andybalaam:
"resolves"? https://github.com/tiimgreen/github-cheat-sheet#closing-issues-via-commit-messages

andybalaam:
I would rename `excludeHidden` to `excludingHidden` and implement it by only `add`ing items we want, rather than copying all of them and `remove`ing the ones we don't want.  Removing from a collection we are iterating through is always painful - I'm sure you are doing it right here, but it's still nasty.

A unit test for `excludingHidden` would be good.

I feel slightly dodgy about excluding items in a different way on different platforms - is that necessary, or could we use `excludingHidden` on Swing too.

Is the text menu missing from this change or did I miss it?

A system test for the text menu would be good too.

andybalaam:
Up to you, but I would not object to some overloads of methods like `info` to avoid passing `false` in lots of places.

andybalaam:
You inspired me to refactor TestAllActiveLevels and amazingly my change doesn't conflict with yours.

andybalaam:
I built a snapshot based on this change and it worked great in Swing but I couldn't get it to work on Android no matter how many times I tapped the rhythm, so I "hacked" it to be always unlocked and uploaded that as snapshot 0.8.0.10-dev-levels in http://artificialworlds.net/rabbit-escape/snapshots/

andybalaam:
Note this was real Android, not an emulator

colonelfazackerley:
Did you see "Mars" in logcat? You have to go back to the first menu, and then forward to the levelset list again.

I should be able to start addressing your comments this evening.

colonelfazackerley:
I don't like resolves for new features. They have several options for bug fixes.

colonelfazackerley:
maybe specifying a "mars" buildType in build.gradle is the way to go for droid.

in text, the dev levels are always available now, but you have to work to get to levels higher than 1. this will need fixing too.

colonelfazackerley:
maybe your droid has timing issues like my emulator. is it elderly?

trying to tap it slowly may help. the detector works independent of tempo.

colonelfazackerley:
or maybe set a property this way

http://tools.android.com/tech-docs/configuration

and read in app with System.getProperty().

I would like to avoid a solution that creates more git fighting bother.

colonelfazackerley:
best i can come up to pre-bake cheatiness into a droid build is to use a name suffix. we get git fighting again, though. boo.

colonelfazackerley:
I managed to unlock cheat mode in my emulator using a slow tempo.

andybalaam:
I wasn't connected to the debugger. I will try again, slower. If it's going to be flaky I'd like to try and think of an additional mechanism.

andybalaam:
It's a Samsung S3. I guess that is elderly now?

andybalaam:
I did the question mark thing again

andybalaam:
A visible indication of cheat mode activation would be really helpful

colonelfazackerley:
I have tried adding ".mars" to the app ID and testing for it. It to work in that the levels are available. However, it crashes without a stack trace showing up in the Debug window in studio when I start a level. I am a bit stuck.

colonelfazackerley:
I think i have it. Appending .mars to the app version. this works for me in debug variant. i tried creating a new variant called mars, but then i get the inexplicable crash when the level starts.

putting .mars in the debug version name does cause git fighting hassle.

colonelfazackerley:
Doing this gives a fixed cheaty apk

```diff
--- a/rabbit-escape-ui-android/app/build.gradle
+++ b/rabbit-escape-ui-android/app/build.gradle
@@ -23,6 +23,9 @@ android {
         release {
             signingConfig signingConfigs.release
         }
+        debug {
+            versionNameSuffix '.mars'
+        }
     }
 }
```

colonelfazackerley:
> I feel slightly dodgy about excluding items in a different way on different platforms - is that necessary, or could we use excludingHidden on Swing too?

The source of your queasiness is that menu building is so different on android. I can't see how to redo this at the moment. It took me ages to see how to do it on android.

swing and text are similar.

please show me how to make text and swing like android.

colonelfazackerley:
I have answered all your points (or made issues to remind me about outstanding ones).

please merge.

andybalaam:
> please show me how to make text and swing like android.

You're probably right.  They are very different, and until we completely refactor Swing (which I may do, because it's a bit of a mess now) they will stay different.

andybalaam:
>Doing this gives a fixed cheaty apk

I'm not inclined to pursue this unless you are keen.  As long as cheat mode is easy to trigger, I don't see a need for a different apk.

andybalaam:
```
> catch ( NullPointerException e )
>        {
>            // This may happen if env var not set.
>            // It can be ignored.
>        }
```

This would be better handled with
```
result = System.getenv( "TAP" );
if ( result != null && result.equals( "mars" ) )
{
  // ...
}
```

Catching expected exceptions is normal in Python, but best avoided in Java.

andybalaam:
Why call `checkEnv` from MenuDefinition in a static block?  Couldn't we just say `public static boolean matched = checkEnv()` inside TapTimer?

andybalaam:
As mentioned above I'd get rid of all this:

`PackageInfo info = getPackageManager().getPackageInfo(getPackageName(), 0 );`

unless you think it is worth pursuing.

andybalaam:
The rest all looks good.

colonelfazackerley:
> Why call checkEnv from MenuDefinition in a static block? Couldn't we just say public static boolean matched = checkEnv() inside TapTimer?

If it's done that way the TapTimer class is loaded later, and matched does not get set to true early. Then you have to play and level and go back to the first menu before it updates to show the dev levels.

colonelfazackerley:
> I'm not inclined to pursue this unless you are keen. As long as cheat mode is easy to trigger, I don't see a need for a different apk.

> I wasn't connected to the debugger. I will try again, slower. If it's going to be flaky I'd like to try and think of an additional mechanism.

Comment conflict. Does not compute.

I am happy tapping out mars on droid. i can get it in a couple of goes. i don' mind if appending .mars to the version is in or out.

Let me know what you want.

colonelfazackerley:
so this is ready to merge, or i can remove droid checking for version suffix first.

andybalaam:
Remove droid checking for version suffix and I will merge immediately. 

I want to find a better way to trigger cheat mode.

colonelfazackerley:
> I will merge immediately.

I started a stopwatch :-)

andybalaam:
Bam

PR #426, opened 2016-05-08
runrabbit text solution player accepts solution strings (coded or not…

colonelfazackerley:
…) as well as solution number

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionDemo.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/solution/SolutionDemo.java
@@ -1,19 +1,17 @@
-package rabbitescape.ui.text;
+package rabbitescape.engine.solution;
 
-import rabbitescape.engine.IgnoreWorldStatsListener;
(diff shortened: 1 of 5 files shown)


colonelfazackerley:
On second thoughts. don't merge yet. I want to make this work on swing too.

andybalaam:
I'd prefer if the solution string came from a different named argument instead of two possible formats for the same argument. But I will merge like this if you want. 

This could do with an update to the docs too.

colonelfazackerley:
I like using single letter abbreviations for the command line arguments. We will run out of letters if we have too many. This would be 3 (solution index, command string and obfuscated command string).

What's wrong with having different formats if they can be differentiated?

Please merge.

andybalaam:
> What's wrong with having different formats if they can be differentiated?

The only rational argument I can come up with is if you accidentally type "1q" when you meant "1" you would get a confusing error message.

colonelfazackerley:
I can change it, if there is a problem.

andybalaam:
Very minor thing

PR #429, opened 2016-05-11
only record successful token drops in swing. closes #414

### LINKED ISSUE #414: Solution recorder records futile clicks
When playing #410 I decided it was necessary to franticly hammer the left mouse button. This resulted in me clicking to try to place more bash tokens than there actually were.  When using the recorded solution as a solution string and running the tests we get:

rabbitescape.engine.solution.SolutionExceptions$UsedRunOutAbility: Solution failed: ability 'bash' was used when there were none left at command 5 of solution 2 in unsorted/LevelItsbehindyou

The solution recorder should ideally not record futile clicks.

colonelfazackerley:


andybalaam:
Nice, ta

PR #432, opened 2016-05-12
water polygons ignore bridges. fixes #430

### LINKED ISSUE #430: Bug: Water polygons don't ignore bridges
I think we decided that water should ignore bridges, so this is wrong.

water-bridge

colonelfazackerley:


andybalaam:
After I merged this it struck me this should have involved either fixing a test or writing a test.

Issue #87, opened 2015-10-10
Level set: "Déjà vu": repeated earlier levels, with higher targets and different solutions


tttppp:
**Levels in this set:**
1. Block drop stop shop #472  **skill:6, thought:7**
2. You wait here but not forever #292 **skill:5, thought:6**
3. Tree, tree, tree again #113 **skill:6, thought:6 [not modified, because I backported the change to #12]**
4. C Red #111 **skill:7, thought:8 [There are 4 ways to do it, so it can't be too hard!]**
5. Oh we do like to be beside the tree-side #160 **[modified]** **skill:8, thought:8**
6. White Flag #235 **skill:7, thought:8**
7. Flagellate #235 **skill: 7, thought:8** [brilliant!]
8. Flagrant disregard #235 **skill:7 thought:8 [excellent fun]**
9. Badtet #158 **skill:8, thought:9 [I found this quite easy (thought:7?) but others found it hard to save more than 7]**
10. Upstairs #57 **skill:7, thought:9**
11. Slippery Slopes" #423 **skill:8, thought:9**
12. Alice in Painland #99  **skill:8, thought:9**
13. Carpal tunnels #70  **skill:8, thought:10**
14. Cups of suffering #18 **skill:9, thought:9**
15. Flagging #235 **skill:9, thought:8 [a bit annoying?]**
16. Tomb of Doom #230 **skill:9, thought:9**
17. The crazy mother #56 **skill:8, thought: 10**
18. Minefield Mind-fail #237 **skill:9, thought:10 [Same as "Carpal tunnels" but save one more]**
19. Three down, one to go #236 **skill:10, thought:9**
20. Keep On Trying #296 **skill:9, thought:9**

**Not included:**
- **[water]** Aquifer **but save 6** #362 (original: ./development/unsorted/Aquifer.rel)
- **[water]** Under pressure **but save 9** #435 (original: ./staging/smallworld/Under-pressure.rel)
- **[no solution]** Gotta catch em all  #75 (development/unsorted/Gotta-catch-em-all.rel)

**TODO**
- ~Read issue comments, create and format level files~
- ~Release a snapshot containing all levels~
- ~Rename any dodgy names (including Bastet)~
- ~Rate for difficulty and re-order~
- ~Update levels with solutions~
- ~Write descriptions~
- ~Write authors and URLs~
- ~Write hints~
- ~Reformat levels~
- ~Ensure music is set in level files~
- ~Add to menu~
[...]

PR #435, opened 2016-05-13
Level: Under pressure

colonelfazackerley:
Look at me: new level by pull req.

I think it's tricky. I am keen to find out if I missed a trivial solution.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/levels/staging/smallworld/Under-pressure.rel
+++ b/rabbit-escape-engine/src/rabbitescape/levels/staging/smallworld/Under-pressure.rel
@@ -1,0 +1,26 @@
+:name=Under pressure
+:description=Careful, that water is highly pressurised.
+:author_name=ColonelFazackerley
+:author_url=https://github.com/colonelfazackerley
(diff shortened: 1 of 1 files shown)


andybalaam:
I will make a snapshot when I get a chance

tttppp:
This seems to cause the tests to fail:

```
Testcase: All_staging_levels_work_and_have_winning_solutions took 0.206 sec
    [junit] 	Caused an ERROR
    [junit] Solution failed: ability 'bash' was used when there were none left at command 28 of solution 1 in smallworld/Under-pressure
```

I assume it was due to https://github.com/andybalaam/rabbit-escape/issues/414 (which is now fixed).

Also I saved 9, is that better than intended?

```
[y#y7$CKgIO|}ggOVy#yy}[IOh'%y[Cy#jOp7yfV#T'CA7AyO~C|IgyzT#yppgOyCK7A%O|}'CpyCy%yy}Op'_gTp#CgKI*O#yyKO#OCVyyf7y'V%yf#CAygy7#y#pCpCIjy#p7py%O~}Ty%7f'~Tj}
```

colonelfazackerley:
yes, your solution is better than mine. it is not simple, though, so I don't mind. 

i suggest leaving 8 as the target and keeping 9 in mind for deja vu.

andybalaam:
#87 deja vu

PR #449, opened 2016-05-18
New level: Infiltrate.

tttppp:
I think this is fairly hard. I've put it in development as it still needs hints.


colonelfazackerley:
> I think this is fairly hard

I would expect nothing less :)

andybalaam:
Still getting around to a snapshot, sorry

PR #485, opened 2017-07-02
Rabbots can't drown

GamingInky:
Made it so rabbots can't drown. (no test yet)


tttppp:
One of these commits looks good! The other one looks like it's part of something else.

andybalaam:
Yeah this needs to be split into 2 PRs

PR #487, opened 2017-07-03
Rabbots can't drown

GamingInky:
There is now a test


andybalaam:
Nice clean diff, nice clean change.

PR #488, opened 2017-07-04
From now on rabbots don't affect the out rabbit count

GamingInky:
That it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,9 @@ levels.txt
 doc/doxygen
 recordings/
 .directory
-/.ropeproject
-.metadata/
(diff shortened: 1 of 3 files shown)


andybalaam:
This needs a test before we can merge. I would also like to look into counting already-out rabbits in the initial count, but that is actually a totally different part of the cod.

andybalaam:
I think there will be a revert step that also needs changing. Make sure the test covers reverting this change too.

GamingInky:
Test: done
Revert step: done
Not repeating: done
Ready for another review!

andybalaam:
Very cool, merged.

Issue #443, opened 2016-05-14
Glitches in water rendering


andybalaam:
Water rendering goes weird in a few places.

Example level:

```
#  P  #
#     #
#######
```
### 1. Sloping water is invisible until there is a certain amount:

![slope-water-invisible](https://cloud.githubusercontent.com/assets/76812/15267101/4660aaca-19b0-11e6-8469-4aff0680de0c.gif)
### 2. Sometimes water level flips between high and low:

![water-level-flips](https://cloud.githubusercontent.com/assets/76812/15267183/5e0d0770-19b2-11e6-8c8f-52570d03b186.gif)

More water level flipping:

![water-level-flips2](https://cloud.githubusercontent.com/assets/76812/15405560/4d1970ce-1dfa-11e6-8952-6eaf9aa1d9ab.gif)

Graph of water levels (see comment for how I got this):

![level-flips](https://cloud.githubusercontent.com/assets/76812/15406752/a657057a-1dff-11e6-8b59-9e906c314a81.png)
### 3. Weird asymmetry:

![water-asymmetry](https://cloud.githubusercontent.com/assets/76812/15405763/353f2a42-1dfb-11e6-8a6e-9c08b5a1dc5c.gif)

Note that when I printed out `wrr.topVertex( LEFT )` and `wrr.topVertex( RIGHT )` during this period, they were identical, implying something else is wrong.

colonelfazackerley:
re 1.
What's happening is that there is no water in the adjacent cell yet. should I render some anyway?

tttppp:
I suspect in 2 the amount of water in the cells is also the problem. It really does flip from high to low sometimes, as the cellular automation happens to resonate with the shape of the world.  Maybe we could create a better visualisation of this, but I think it would be hard. The polygon representation is much nicer than the earlier sprite version.

We could consider plotting the height of water at the centre of cells, rather than using the edges. This would result in drawing water in empty cells sometimes, but might be smoother to animate. I think we will get problems with full cells next to empty cells (they won't look good).

andybalaam:
I want to get a bit more info to check @tttppp's suspicion. I will also add a couple of other gifs. I suspect there may be a couple of actual rendering bugs but could be wrong

andybalaam:
Having lines bending in the middle of a square might make it harder to tell when a rabbit was going to down

tttppp:
@andybalaam Yes - agree about hard to see drowning.

[...]

PR #538, opened 2018-08-08
Water rendering

tttppp:
Not perfect, but the bugs from https://github.com/andybalaam/rabbit-escape/issues/443 are much improved.


andybalaam:
Snapshot 0.10.3.5 contains this code, plus my copy of the Swing code into Android: http://artificialworlds.net/rabbit-escape/snapshots/

andybalaam:
screenshot_20180812-020259

Working nicely with one more fix, in snapshot 0.10.3.6

PR #558, opened 2018-08-17
Fix rabbot brollychuting animation - fixes #555.

### LINKED ISSUE #555: Bug: Animation for brollybot is missing
For example give the rabbot a brolly token on One From Each Side, and make it dig down. The rabbot changes to white when it pulls out the umbrella.

Tested on my laptop and a recent Android snapshot.

tttppp:


tttppp:
For #555

andybalaam:
Awesome, I found this on the train this morning, and also noticed we are missing sound effects, so I am working on that part.

PR #559, opened 2018-08-17
Add hints for three rabbot levels.

tttppp:
Add remaining hints for all my levels except "* close for comfort".


tttppp:
I couldn't think what to write for those levels since they're so small.  If you can come up with something then please go for it!

andybalaam:
Rebased and merged, thanks!

PR #562, opened 2018-08-19
Fix older level names to contain actual names.

tttppp:
Use horrific bash command to rename all old fashioned levels:

```
for f in `find rabbit-escape-engine/src/rabbitescape/levels/ -name "level*.rel"`; do line=`cat $f | grep ":name"`; name=`echo $line | sed "s|:name=||g"`; if [ "#$name" != "#" ]; then dir=`echo $f | sed 's|[^/]*$||g'`; number=`echo $f | sed "s|.*level_\([0-9]*\).rel|\1|g"`; g=`echo ${number}_${name}.rel | sed 's| |-|g' | sed 's|[^ 0-9a-zA-Z_.-]||g'`; mv $f $dir$g; fi; done
```


andybalaam:
Nice, thanks.

PR #569, opened 2018-09-07
WIP: Experimental dirt blocks

doctormo:
This branch contains some designs for creating a different block feel for the game. I understand though if the original game likes the crayon style, I wanted to try out an svg based style using elements from openclipart.

screenshot rabbit

The desired outcome is to have grass on each of the blocks and a shadow on each underside, this would require some code to produce a new graphical layer which is painted on top of and offset from the game layer.


tttppp:
I like this, perhaps as a random level-by-level selection (using the same system as the paper edge design).

The exit looks like a picnic table to me at the moment. It would be interesting to see a screenshot of a level with metal, and perhaps water too.

The only other concern I have is the resulting timewise, but presumably this isn't as bad as e.g. adding a new rabbit behavior or a new mp3 snippet.

tttppp:
*timewise --> filesize

andybalaam:
Wow, this looks great!  File size should be fine.  The question I guess is: do we like this better than the current design, or do we want to support both?

andybalaam:
btw the idea of painting another layer on top (higher than the rabbits) is definitely in line with what I expected we would need some day.

andybalaam:
@colonelfazackerley @GamingInky what do you think?  I am quite excited by how that screenshot looks.  I don't have time yet to review the actual change.

doctormo:
It's great to hear a positive reception.

The `picnic` table is actually a Japanese gate. It's actually surprisingly hard to think of ways for rabbits to enter and exit the world.

tttppp:
I think having multiple style, but fixed per level, is a nice idea. Lots of games have used this to add a sense of adventuring through different places. E.g. Lemmings and Worms spring to mind.

The current dirt design looks interesting enough to me. We could maybe add occasional skeletons or buried treasure, but I think that could be a future addition.

doctormo:
Oooh, I like the ideas. If we had a terracotta pot as the starting point, between that and the garden gate we'd give it a real peter rabbit feel ;-)

I love the idea of a random skeleton. Although we can't currently specify the rarity of a block right.

andybalaam:
Loving the flowerpot idea.  Yeah, we can't specify the rarity of a block but I imagine it could be added pretty easily.

andybalaam:
Maybe different level sets could use different block styles, like the music works now.  Then we could have all kinds of exciting looks and feels.

tttppp:
I was thinking that a level creator could specify the tile set for their level, or it would be deterministicly random if they didn't specify. This would allow deliberate use of the theme, e.g. linking the title to the visuals. (E.g. Titles linked with Peter Rabbit is a great example)

andybalaam:
Sounds good.  Music works like this, except we don't choose an arbitrary one if you don't specify.  (I'm not too bothered about choosing deterministically-at-random, since updating the released levels to specify seems pretty easy.)

colonelfazackerley:
The dirst blocks look nice, but somehow I think i prefer the game without them. I think it's because it makes the games look more like other games. If you search on google play for rabbit escape there are a few games. The papery aesthetic of this one makes it unique. 

The graphics can definitely be improved, but I think going for a more realistic (if still cartoon) style would be a change in the wrong direction.

andybalaam:
The most important reason I like the screenshot above because it is colourful. At the moment the game o is very monochrome. If we could keep the paper, hand drawn feel, but make it more colourful, that would be brilliant.

colonelfazackerley:
might be cool to have colourful dirt blocks, grass etc, but with a hand drawn look.

andybalaam:
Definitely worth a try.

andybalaam:
Something like this? 
image

andybalaam:
(Obviously without grass in the middle of the ground, and with grass in front of rabbits, and with sloping blocks done too, and with flowers actually visible.)

colonelfazackerley:
I like it. I might try a bit of transparency to see if it makes the scribble texture a bit less aggressive.

andybalaam:
I pushed my branch here in case it's helpful https://github.com/andybalaam/rabbit-escape/tree/wip-brown-earth

andybalaam:
It was 5 minutes' work, so I have no emotional attachment to it :-)

doctormo:
There's also a paper texture bitmap effect you can apply to svg objects, it might provide the right kind of texture for the blocks.

doctormo:
Addendum: Shadows are really important for separating out each of the layers, so elements that should layer on top of each other should somehow contain them. (although I admit that might be a rendering burden)

andybalaam:
Shadows sound good, and at least on Android I don't think we have a performance problem - I would expect semitransparent pngs to be fine.  (We convert all our svgs to png for now, since the oldest Android versions we support don't have an svg rendering library.)

andybalaam:
This is a great idea, and I have opened #602 to track work on it.  This PR was never intended to be merged, so closing.

PR #572, opened 2018-09-15
list dependencies for fedora

### LINKED ISSUE #570: Rabbits on Fedora
most things on INSTALL.md are the same.
`sudo dnf install git make ant sox inkscape expect grep sed python3-lxml`

couple of things have different package names
`sudo dnf install java-1.8.0-openjdk-devel ImageMagick`

I can't find a fedora package "rename", or what rabbit escape uses from ubuntu's rename. I carried on without it. I do have a standard linux util called "rename" (from a fedora package called util-linux). 

INSTALL.md says to `make all-images` first. That failed. `make compile` first seems OK. `make all-images`after gives no errors, but I think is not required now.

I could do a pull request for INSTALL.md adding the fedora dependencies, and removing the bit about `make all-images`, but that is possibly a makefile bug better fixed. "rename" thing probably needs clarifying.

My machine has functioning sound, but I'm not getting any from rabbit escape.

(
I'm trying fedora because a buggy BIOS issue stops ubuntu from working.
https://askubuntu.com/questions/1073434/acpi-errors-prevent-boot-after-upgrade-to-kubuntu-18-04/1074424#
)

colonelfazackerley:
closes #570

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -40,6 +40,11 @@ to build the images, sounds and code.  On Ubuntu, Debian and similar:
 Make sure you have Inkscape 0.92 or above to avoid slightly-wrongly-sized
 images being generated.
 
+On Fedora some of the packages have different names.
+
+    sudo dnf install git make ant sox inkscape expect grep sed python3-lxml \
+        java-1.8.0-openjdk-devel ImageMagick util-linux
+
 Get the source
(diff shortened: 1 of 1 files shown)


andybalaam:
Awesome, thanks.

Issue #421, opened 2016-05-07
GUI feedback for successfully tapping rhythm... 


colonelfazackerley:
...to enter cheat mode.

Issue #382, opened 2016-03-29
Level: Escape from Shurrapak


colonelfazackerley:
I think this is hard. Possibly not in a good way.

It started from The traitor, but diverged a lot. It's too different for deja. Not sure where this will fit.

```
:name=Escape from Shurrapak
:description=[To Gilgamesh] I thought you were dead.
:author_name=colonelfazackerley
:author_url=https://github.com/colonelfazackerley
:hint.1=
:hint.2=
:hint.3=
:solution.1.code=7[ce^dF7[-*^de7[c*^de0c,^Gc,^_20-7[-7[-m[-^F7m[c^F7m[c^F7m[-^-[c[c7c[F[-[-fmdF^F7[-e7dF*[-*^de7[^*^97[7<7*c9x9Fef,^dF/ed_7d9*^0[72eFx^^c[e77cd*7[^e'ed^^[--777*d*c[^e'ee^^dc7-77-d,[[[d'd0^^-c^-77Fdd2[d['[c^ed-^/7-dd7e[2d'e-^[/cf^7^eFc6
:music=tryad-beat_into_submission
:num_rabbits=5
:num_to_save=5
:bridge=30
:bash=5
:dig=4
:rabbit_delay=9,6,5
   P               

  P     #        # 
###   #   #     # #
#Q    #  (  # #    
   #  # ### #P ## O
##(  ( ( ( (  ((( #
###################
```

andybalaam:
Looking forward to trying these. Finding myself wanting to be able to try development levels in snapshots. Maybe in cheat mode, or in a second secret mode.

colonelfazackerley:
you mean on android?

in swing i use the load level feature and navigate to the dev levels under blah..engine..blah..blah..blahdey...blah...blah.

colonelfazackerley:
seeing the merge comment in #383 

perhaps this could be an extra for outdoors.

andybalaam:
Sounds good.

andybalaam:
Yes, I do mean Android. I do most of my rabbit escape work on the toilet these days.

andybalaam:
I am working on https://github.com/andybalaam/rabbit-escape-level-editor/issues/1 so we can put this level on the forum.

andybalaam:
Now the forum supports it, moving this to here: https://www.artificialworlds.net/rabbit-escape/levels/d/28-escape-from-shurrapak

Issue #367, opened 2016-03-15
Enable cheat mode via command line option


andybalaam:
Maybe a command line option to enable cheat mode, to make it easier in development?

andybalaam:
Adding environment variable TAP=Mars seems fine.

PR #578, opened 2018-10-16
particles for falling water

colonelfazackerley:
still WIP. gif in a mo.


colonelfazackerley:
re

andybalaam:
This looks super-cool.  Did you turn off the blue background colour for falling water?  I think I'd like it to be there, but more subtle.

colonelfazackerley:
I put a shaded water background back.
re

@tttppp note that you dont need to iterate the 2D table. LookupTable2D implements iterator, you can just  iterate the objects in it. You didnt need to iterate, x, y and skip the nulls.

andybalaam:
I am liking this.  What do others think?

tttppp:
I like this! I think the performance might be a potential issue (I'm fairly sure the colonel raised this point himself in the past).

I think the background colour is particularly useful to show cells that contain a high density of water. It helps to avoid the 'jump' between 1024 units falling and 1024 units standing still.

Obviously stopping the water at solid blocks would be cool. Could this be done by pseudorandomly generating the drop location based on (tick+y coordinate) as the seed?

colonelfazackerley:
I think your suggestion would be very different to my current implementation. I have some more ideas for improving it along the lines of how the PR works now. I'd like to try those first.

re: performance. I'm surprised how fast it runs at the moment. I think I will keep going and run a profiler if I have to. I also have an idea to draw fewer drops on struggling machines, but that's probably thinking so far ahead it's not useful.

colonelfazackerley:
particles no longer fall through full cells (including solid blocks).

re

tttppp:
Looks really good to me! Nice work!

colonelfazackerley:
water sprays out from pipes. still more tweaks in mind.
re

andybalaam:
Wow, this is looking so good!  Is the performance OK?  On low-powered computers (e.g. Raspberry Pi) can it back off somehow?

andybalaam:
I see drops flowing through that bottom block btw...

tttppp:
Very cool! You should try a slope in your example level, as I think that's likely to cause issues if the raindrops are 'in front' of the dirt.

tttppp:
Also I couldn't determine if the dampening causes the drops to have a maximum velocity or not. A maximum velocity would be useful with the current system to prevent drops falling through a block of dirt and appearing in the block below.

Btw I just realised that drops are behind dirt, but dirt is transparent.

colonelfazackerley:
perfomance: it's not making the fan on my computer come one. in previous JVMs the standing water polygons did. I did this hack:
```
diff --git a/rabbit-escape-render/src/rabbitescape/render/gameloop/GameLoop.java b/rabbit-escape-render/src/rabbitescape/render/gameloop/GameLoop.java
index 16855662..e148089f 100644
--- a/rabbit-escape-render/src/rabbitescape/render/gameloop/GameLoop.java
+++ b/rabbit-escape-render/src/rabbitescape/render/gameloop/GameLoop.java
@@ -109,7 +109,7 @@ public class GameLoop
         long wait_time = frame_time_ms - how_long_we_took;
 
         input.waitMs( wait_time );
-
+        System.out.println( " " + wait_time );
         return input.timeNow();
     }
```
when playing escape from shurrapak (#382) i'm getting 60-70 of 100ms waits. You can tone the droplets down by tweaking WaterRegionRenderer.contentsPerParticle. 1 gives lots of particles, 4 is current, before 1000, it turns it off. Between 4 and 1000, I am not seeing much change in the wait_time. so, lets fuss about it when we have to. if my laptop fan comes on i can check the last diff.

there are drops in the solid blocks, but they dont come out below. i think this is ok. the block just looks a bit wet.

drops do have a terminal velocity. I think it's equal to gravity/(1-damping). with current values
- gravity 0.008 cells per anim step squared
- damping 0.95 

this comes to 0.16 cells per animation step. I leave derivation of the formula as an exercise for the reader.

damping is a bad name for that variable, as it represents the velocity retained, not the velocity lost. cant think of a better one though.

colonelfazackerley:
particles init velocity influenced by net flow.

re

tttppp:
Velocity retained might be called "momentum" in common usage (although scientifically that's a different type of measure).

Looks good to me now, even with a small amount of overlap with the dirt.

andybalaam:
This looks amazing, and has got me all excited about releasing some water levels.

colonelfazackerley:
As middle aged men, we will soon all be excited about releasing water.

colonelfazackerley:
I'm quite pleased with it so far. I think I can improve it to. Play #382

andybalaam:
Cool, let me know when you consider it ready for review.

andybalaam:
Yes!

colonelfazackerley:
After the colour varying, I thought the particles in the same cells as the solid blocks looked worse, so I reduced it.
re

andybalaam:
Not sure I'm liking the colour varying.  Different darknesses might work, but different colours looks a bit unwaterlike to my eye.

colonelfazackerley:
I'm not decided on drops of differing hue. I can try tonal difference only, or perhaps differing max opacity. I gave the streaks width. I left them quite fat for this gif. We can slim them down easily. Though the fatness gives them a cartoony style, which might be good.


re

colonelfazackerley:
They already fade in and out using opacity, so it's not going to decrease performance. I take your point about the standing water being uniform. Giving it some texture to match feels like a step too far at this point.

andybalaam:
I suspected so on opacity.  Yeah, also, I like the uniformity of the standing water, I think.

colonelfazackerley:
opacity varies now. i think i prefer it.
re

andybalaam:
Yeah, really like it.

andybalaam:
That poor rabbit.

colonelfazackerley:
@tttppp I tried slopes. I'm not sure I understand your point: "You should try a slope in your example level, as I think that's likely to cause issues if the raindrops are 'in front' of the dirt."
re

tttppp:
No - that example is brilliant. Thanks!

My point was wrong. I tried to retract it when I said:
> Btw I just realised that drops are behind dirt, but dirt is transparent.

However I think it's great to see another example, particularly with droplets entering from several sides of a cell.

andybalaam:
The slopes look great.

colonelfazackerley:
I did notice that the background shading is lighter than it should be for solid slopes (bridges are OK). I think I understand why. I will fix that.

colonelfazackerley:
I rained harder on the poor rabbit. Mwahahaha. It drops the number of particles if the computer is struggling. I made a level with loads of particles to test. GIF is too big, but here are some stills.

At the start, the rain has not got far, so each drop represents less water.
frame_0000040

At the end there is a lot of rain on the screen, so it has turned up the amount of water each drop represents, reducing the particle count.
frame_0000226

colonelfazackerley:
just android to sort now, I think. (see the todo list in post 2, which I have been updating).

andybalaam:
The latest gifs made me laugh out loud.  Really glad the low-hardware change is in.  Makes me even more confident about this.

colonelfazackerley:
Particles on Android.
deepinscreenshot_select-area_20181109095148

colonelfazackerley:
The crash was on swing too. Fixed now. The "development" level "Waterfalls" is pretty. I think it uses the particle count moderation code on my tablet. This seems to work well on android too.

andybalaam:
Amazing!  Can we stress-test this somehow to check for more crashes?  Maybe just run it for a few weeks on our own devices, while we develop water levels.

colonelfazackerley:
I tried to test on API level 8. Android Studio says it has timed out waiting for the virtual device to start. The virtual device looks started. Not sure how to fix. Perhaps @andybalaam can test on API level 8.

As far as I can see this can be merged.

colonelfazackerley:
API 16 is the oldest virtual device I can get to work. It doesn't crash, but it's unplayable, I thin due to the inefficiency of the virtual machine.

colonelfazackerley:
Nexus tablet Android 7.1.1: works very well on Waterfalls
Pixel phone Android 9: works OK on Waterfalls, but does moderate the particle count down a lot occasionally for reasons that are not clear to me.

colonelfazackerley:
After getting the apk onto my phone, I did a profiling session (developer menu). This makes a .ctrace file (I think thie requires API 28), which I emailed to myself. I converted this to html with
```
cd ~/Android/Sdk/platform-tools/systrace/
python systrace.py --from-file ~/Downloads/trace-sailfish-PPR2.181005.003-2018-11-09-21-46-26.ctrace -o ~/Downloads/trace.html
```
I was hoping to be able to see if the water stuff was taking significant time compared to other bits of code. I could not figure it out.

tttppp:
This is mega-exciting! How can I test water particles on my device? Do I need to use the cheat mode? (I don't think I've ever got that to work)

colonelfazackerley:
Android studio is one way. I tap mars, and it works for me. Can you build your own from this branch? You could replace normal levels with watery ones if you don't have the rhythm. Perhaps @andybalaam could put up an apk/jar of this dev branch, I think he's done that before.

colonelfazackerley:
@tttppp I made a version of rabbit escape with 10 water levels as a set for testing. The build is producing split APKs. I could upload them somewhere for you, but I have no idea how you would install them. I think we need @andybalaam for advice.

colonelfazackerley:
@tttppp I think I figured it out. turning off "instant run" stops the splitting. Here is a watery APK

https://drive.google.com/open?id=1gA9giZFiKGvEgoPJae-y7fiORGQGpZFH

andybalaam:
Related: #421 #367 (is that done, making my comment above wrong?)

colonelfazackerley:
re: sidenote. I can usually get Mars the first time. On some devices it takes ages: I think this is the device's timing being off. There is a command line method for enabling Mars, but only for the text ui :). For swing or android (if you are USB debugging) it prints 'Mars' to the console. 

For an alternative method of triggering how about dropping a bridge token directly under the entrance on arcade 1 (ghost vs pie)? It would have to be the first rabbit, and he would have to die off the top of the world on a specific square. I think this meets the criteria of being easy if you know, but unlikely to be found by accident. if @andybalaam agrees I will make a PR. I would remove the mars timing code in the same PR.

andybalaam:
I made a forum post: [Ghost vs pie](https://www.artificialworlds.net/rabbit-escape/levels/d/42-ghost-versus-pie) to help visualise this.

I suggest dropping the bridge token on the far-right square, so it is easier to do without very fast reactions.

andybalaam:
By the way, we don't need to delete Mars tapping - it's a nice fun thing.

andybalaam:
Oops, I didn't mean far-right square.  I illustrated a better suggestion here: https://www.artificialworlds.net/rabbit-escape/levels/d/43-proposed-cheat-mode

colonelfazackerley:
It's the rightmost square of the pie's flat top. I understood.

rabbit-escape-engine/src/rabbitescape/engine/WaterRegion.java
```diff
@@ -104,7 +104,10 @@ public boolean isConnected( CellularDirection direction )
 
     /**
      * Get the amount of water being transferred from here in a given direction
-     * this tick.
+     * this tick. Note that it is always posotive. For example if water is
```
andybalaam:
```suggestion
     * this tick. Note that it is always positive. For example if water is
```

rabbit-escape-engine/src/rabbitescape/engine/WaterRegion.java
```diff
@@ -104,7 +104,10 @@ public boolean isConnected( CellularDirection direction )
...
+     * flowing into a cell from the left, it's LEFT will give zero.
```
andybalaam:
```suggestion
     * flowing into a cell from the left, its LEFT will give zero.
```

rabbit-escape-engine/src/rabbitescape/engine/WaterRegion.java
```diff
@@ -104,7 +104,10 @@ public boolean isConnected( CellularDirection direction )
...
+     * However the cell to it's LEFT will yield a posotive number on it's
```
andybalaam:
```suggestion
     * However the cell to its LEFT will yield a positive number on its
```

rabbit-escape-engine/src/rabbitescape/engine/config/StandardConfigSchema.java
```diff
@@ -20,6 +20,18 @@ public static void setSchema( ConfigSchema definition )
...
+            String.valueOf( true ),
+            "Particle count is auto-reduced for struggling hardware."
+        );
+
+        definition.set(
+            CFG_WATER_CONTENTS_PER_PARTICLE,
+            String.valueOf( 4 ),
+            "Smaller values lead to more particles"
+        );
     }
```
andybalaam: I think we need a new config version, since we're adding config properties.  The upgrade will probably be empty, but should set some default values if it is not OK for these values to be empty.

Alternatively, do we really need this to be configurable?  If not, I'd be fine with hard-coding these in the source code.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -85,7 +104,33 @@ else if ( isSlopeNotBridge( block ) )
     }
 
     /**
-     * Called once per animation step
+     * Fades splash particles in/out to approximate
+     * correct anount of falling water. Called once per game step
```
andybalaam:
```suggestion
     * correct amount of falling water. Called once per game step
```

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -95,6 +140,83 @@ public void setWaterHeight()
...
+            int heightInCell = (int)( (1.0f + Math.floor(p.y) - p.y) * 32.0f );
+            if ( heightInCell < r.targetWaterHeight )
+            { // below water level: delete immediately
+                particles.remove(p);
+            }
+        }
+    }
+
+    /**
+     * Updates particle position. Called onces per animation step.
```
andybalaam:
```suggestion
     * Updates particle position. Called once per animation step.
```

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -288,6 +410,61 @@ boolean isFull( int x, int y )
...
+            if ( t instanceof Pipe )
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns the ney flow across a cell's edge considering water flowing
```
andybalaam:
```suggestion
     * Returns the net flow across a cell's edge considering water flowing
```

rabbit-escape-engine/src/rabbitescape/engine/config/StandardConfigSchema.java
colonelfazackerley: I think I would like them to be configurable. I don't understand how the config versioning works. Can you explain? Or make the change?

andybalaam: Why would you like them to be configurable?  Surely we just want to find the right values and set them?  Or do you see us shipping different builds with different config?  Or do you see the need for users to modify them?

andybalaam: Some info on config upgrades: there is a list of how to upgrade from one version to another here: https://github.com/andybalaam/rabbit-escape/blob/master/rabbit-escape-engine/src/rabbitescape/engine/config/RealConfigUpgrades.java , and then the actual upgrades are in e.g. here: https://github.com/andybalaam/rabbit-escape/blob/master/rabbit-escape-engine/src/rabbitescape/engine/config/upgrades/UpgradeTo02LevelsCompletedByName.java

Essentially, you increase the config version number by 1, and make a new upgrader that converts from the previous version to the new one.

Bear in mind though that if you allow these settings to be missing and provide a default, we probably don't need a new version at all.  I don't remember, but I think that is true.

andybalaam:
Unless I missed something, we're waiting for an answer from @colonelfazackerley on whether the number of particles needs to be configurable.

colonelfazackerley:
I did test this snapshot on multiple droids
https://drive.google.com/open?id=1gA9giZFiKGvEgoPJae-y7fiORGQGpZFH
 
Also - sorry about the typos in the comments.

andybalaam:
Np, I love spotting typos - makes me feel useful ;-)

andybalaam:
Snapshot 0.11.0.1 contains this! http://artificialworlds.net/rabbit-escape/snapshots/

andybalaam:
Waterfalls works pretty smoothly on my not-very-fast Moto C.

andybalaam:
And it looks great.

PR #586, opened 2018-11-12
another route to mars mode

colonelfazackerley:
another route to mars mode, as discussed in #578.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/behaviours/OutOfBounds.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/behaviours/OutOfBounds.java
@@ -7,6 +7,7 @@ import rabbitescape.engine.BehaviourTools;
 import rabbitescape.engine.ChangeDescription.State;
 import rabbitescape.engine.Rabbit;
 import rabbitescape.engine.World;
+import rabbitescape.engine.config.TapTimer;
 
 public class OutOfBounds extends Behaviour
 {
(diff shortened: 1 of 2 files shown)


andybalaam:
Nice, thanks!

tttppp:
Thanks! This is great for those of us without an intricate knowledge of Holst.

I thought it was worth noting that it's not exactly an insane thing for someone to do by accident. (e.g. crazier would be dropping an explode token there!)  So I have a slight objection to this being discovered by accident and there being no feedback - someone might think this is a bug rather than a feature.

colonelfazackerley:
I would have slightly preferred the token to be under the exit (less likely to be done by accident), but I'm OK with where it is. Blowing a rabbit up in a random spot would have worked too. I wouldn't mind a more explicit notification of entering mars mode. Laziness just beats interest for me on that point. If I could think of a super low-effort way to implement it, I would.

andybalaam:
I'd be totally happy with dropping an explode token there.  I didn't want it under the exit because it requires fast fingers and stresses me out.

A visual indication would be nice: to prevent people thinking they'd found a bug, it would need to be a dialog box that popped up, similar to when you finish a level.

andybalaam:
Verified this works for me, and is much easier than tapping!

tttppp:
Same for me.  Thanks for this!

andybalaam:
image

PR #589, opened 2018-11-19
fix expect tests: levels have names in filenames

### LINKED ISSUE #584: slowtest fails
After #583 ```make slowtest``` fails. Something to do with ```expect``` and the text UI.

colonelfazackerley:
fixes #584 

the android smoketest still fails after getting quite a way through, I think. this improves things.


andybalaam:
Nice, thank you!

PR #590, opened 2018-11-21
Bash block dig bug fix

### LINKED ISSUE #587: bash-block-dig bug(s)
```
:name=bug
:bash=1
:block=1
:dig=1
:solution.1=bash;(1,1)&block&(1,1);dig;2;(1,0);4
r #
 ( 
#  
```
Rabbit standing on bridge bashes, but is turned into blocker before moving into the cell from which the block was bashed. Rabbit is then turned into digger.
1. The rabbit as blocker does not stand on the bridge properly
2. The memory of being a basher is enough for them to walk on air.
re
I think this is a bug that has been fixed before, but not for all cases.

colonelfazackerley:
fixes #587 

I believe changing the rabbit position ```++t.rabbit.y``` in ```newState``` is naughty. I tried changing it in ```behave```, but you get at least 1 anim frame in the wrong place. This fix is simple and works. I don't mind changing it, if there is a more correct solution.

@tttppp mentioned that this might break levels on the forum. I have not managed to complete them all. If they rely on bugs, I think they should be changed.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/BehaviourTools.java
@@ -82,6 +82,24 @@ public class BehaviourTools
         }
     }
 
+    public boolean rabbitIsBashing()
+    {
+        switch( rabbit.state)
+        {
+        case RABBIT_BASHING_RIGHT:
+        case RABBIT_BASHING_LEFT:
+        case RABBIT_BASHING_UP_RIGHT:
+        case RABBIT_BASHING_UP_LEFT:
+        case RABBIT_BASHING_USELESSLY_RIGHT:
+        case RABBIT_BASHING_USELESSLY_LEFT:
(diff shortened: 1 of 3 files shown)


colonelfazackerley:
I have not managed to complete all the levels on the forum, and I dont want to spoil them for myself. If you  find a specific undesirable effect of this fix, please make a mini level like I did in #587 to demonstrate it.

tttppp:
I've made a start migrating the missing levels from the forum: https://github.com/tttppp/rabbit-escape/tree/ForumLevels
I've just added the missing water levels so far.

andybalaam:
Brilliant, thank you.

andybalaam:
OK, if I understand your test correctly, it should still pass even if we find a better fix, so this looks good to merge.  Thanks for finding and fixing this!

andybalaam:
Snapshot 0.11.0.2 contains this: http://artificialworlds.net/rabbit-escape/snapshots/

PR #592, opened 2018-11-24
remove unused imports and unused variables. fix a bug in water partic…

colonelfazackerley:
…les where splashes might go the wrong way. found from checking unused variable.

I was using a text editor. I set up eclipse, and found a bunch of things that were my fault.


andybalaam:
Would be great if we had some kind of continuous integration telling us things like this.

colonelfazackerley:
continuous integration can be done through github. I think that would have to be you. Would you prefer to do it outside github?

andybalaam:
Inside github is fine - I would just need to get around to it.  Really we should have a shared Rabbit Escape area which would mean others could help.  While we're at it we could consider moving to GitLab since it's more Open Source.  GitLab has some nice CI options too.

PR #593, opened 2018-11-24
Forum levels

tttppp:
Here are all the levels (and extra solutions) that I could find in the forum.  This should allow us to see the impact of changes on existing levels, and decide whether e.g. the intended solution of a level exploits a bug.


andybalaam:
Very useful, thank you.  In future, when Rabbit Escape can see the levels in the forum and play them directly, we won't need to do this (we can have a Makefile target that just reports on the playability of all forum levels) but for now it is great, thanks!

andybalaam:
Snapshot 0.11.0.2 contains this: http://artificialworlds.net/rabbit-escape/snapshots/

colonelfazackerley:
I called by recent level "Nice cup of tea" to avoid calling it "cup half empty". I totally forgot/never knew there was already a "cup of tea" level...

andybalaam:
Cup of tea is also new, so I may have copied you.  I spontaneously made it during the filming of the level editor video :-)

PR #595, opened 2018-11-25
Two more water levels.

tttppp:
I previously missed "Escape from Shurrapak", and this also adds "Burst water main".


andybalaam:
Snapshot 0.11.0.3 has these: http://artificialworlds.net/rabbit-escape/snapshots/

PR #596, opened 2018-11-27
Fix bash from bridge to dig again

### LINKED ISSUE #591: bashing from bridge -> digging bugs
```rel
  
r*
##
  
:bridge=1
:block=1
:*=(bd
:solution.1=bridge;4;(1,2);4
:solution.2=block;4;(1,2);4
```
bashing from a bridge and then transitioning to digging seems problematic. I suspect, like #590 the solution may be to drop the rabbit back onto the slope before they start digging.

colonelfazackerley:
fixes #591.

it's a redo of #590, but covers #587 as well as #591.

It still changes rabbit coordinates outside of behave(). I couldn't make that work. I tried to put the code in Bashing, which is the cause of the problem, but could not. It had to be in Blocking and Digging.

The minimal change I came up with did result in a timing change which broke "Infiltrate" and "Under pressure". I submit new solutions for those. @tttppp might like to make another solution for "Infiltrate" if this PR is accepted.

### DIFF
--- a/rabbit-escape-engine/src/rabbitescape/engine/Rabbit.java
+++ b/rabbit-escape-engine/src/rabbitescape/engine/Rabbit.java
@@ -28,6 +28,9 @@ public class Rabbit extends Thing implements Comparable<Rabbit>
 
     public Direction dir;
     public boolean onSlope;
+    /** Rabbits move up 1 cell to bash from a slope.
+     *  Keep a note, so it can be undone.  */
+    public boolean slopeBashHop = false;
     public final Type type;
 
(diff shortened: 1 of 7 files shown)


andybalaam:
Nice work, thank you.

But: yuck!  Do we need a new ability called BashingOnSlope?

tttppp:
I think part of the reason Infiltrate was so hard was this bug. I guess that's why the original solution.2 was written.  I haven't solved this with the timing changes, but I expect @colonelfazackerley's solution is a fine replacement for solution.1.

PR #598, opened 2018-12-01
80 char line limit.

colonelfazackerley:


andybalaam:
I don't like this:

```java
    public static List<WaterRegion> makeWaterRegion( int x, int y,
                                                     Shape[] shapes,
                                                     boolean outsideWorld )
```

I would prefer this:

```java
    public static List<WaterRegion> makeWaterRegion(
        int x,
        int y,
        Shape[] shapes,
        boolean outsideWorld
    )
```

I like to "chop down", so if we flow onto the next line I want to start at 1 indentation in, not lined up with the opening bracket.  Justification: renaming makeWaterRegion won't mess up this formatting.

If we have to flow onto multiple lines I line each argument to get its own line.  Justification: It is easy to miss arguments if they are on the same line: it makes some arguments more prominent than others.

Generally my preferences match Kevlin's quite closely: https://vimeo.com/74316116

andybalaam:
Thank you for doing this work, by the way.

colonelfazackerley:
I made some adjustments. Probably didn't get everything, but better than it was.
@andybalaam

andybalaam:
There are still quite a few not chopped down e.g. createWaterRegionsAtPoint, chooseState, switchType, baseVariantSwitch

Are you doing this manually?  Maybe this should wait until we have a decent automated tool?

colonelfazackerley:
@andybalaam  I fixed a bunch more. Yes, this is handcrufted. I should probably have used an auto tool, but none seems quite right. I would quite like to get this finished.

I rebased the tidy commits into the the first.

andybalaam:
Thank you for this.  It's definitely better than before, with more room for improvement!

Issue #605, opened 2019-01-10
Bridges make water think it is falling


andybalaam:
![image](https://user-images.githubusercontent.com/76812/50986203-1189c100-14fe-11e9-8ffc-af30112994ea.png)

A bridge below seems to make water think it is falling when it shouldn't be.  I'd expect the above screenshot to show a full-looking cup of tea.

colonelfazackerley:
Are the cells full? If you pause the overlay shows the amount of water in each cell. Knowing this will help to show if the bug is in engine or render.

andybalaam:
Really enjoyed generating this screenshot because water is so cool.

andybalaam:
![image](https://user-images.githubusercontent.com/76812/50996790-ccc05300-151a-11e9-94c6-2ab23335add0.png)

Yep, definitely looks like rendering is the problem.

colonelfazackerley:
I've had a few goes at making a simple level that I can just run without user interaction to demonstrate the problem. I can't reproduce it. Can you post a simple level, or a solution to a level that shows the problem?

andybalaam:
Interesting.  When I make the bridges be there at the start of the level it doesn't seem to happen...

andybalaam:
Ah, got one:

```rel
#P  i #
#     #
#    (#
#     #
#r    #
#iri  #
#######
```

andybalaam:
![tmp](https://user-images.githubusercontent.com/76812/51027087-eba8fe80-1587-11e9-9818-8add776fa4f3.gif)
I love just watching water flowing.

colonelfazackerley:
isFalling() looks uses isFull() for the cell below. isFull needs the height and targetWaterHeight to reach 32. In the cell with the bridge in above the solid block these get stuck at 19. WaterRegion.getContents() returns  622 (622/32=19). Now need to figure out why that is stuck. Problem actually seems to be in engine...
@tttppp ?

colonelfazackerley:
in setTargetWaterHeight() WaterRegionRendered.getContents() is used, and returns 622 for the cell. When the overlay is drawn WaterRegionRenderer.contents is 1616. Need to understand why this number is changing within 1 time step.

tttppp:
If it's a problem in the engine then it'll be because bridges are being treated like slopes for some reason. We should fix this before releasing a load of levels with solutions that depend on the current behaviour :-)

tttppp:
[...]

PR #611, opened 2019-02-24
Remove state from renderer. Rely on world for state.

tttppp:
This fixes !605 - water displayed as falling even when cells below are full.

Also some wildcard imports, which IntelliJ insists are mandatory (and so
presumably they are desired based on the settings in the .idea directory).

### DIFF
--- a/rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
+++ b/rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
@@ -1,36 +1,22 @@
 package rabbitescape.render;
 
+import rabbitescape.engine.*;
+import rabbitescape.engine.textworld.BlockRenderer;
+import rabbitescape.engine.util.*;
+import rabbitescape.render.gameloop.WaterAnimation;
+
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
-import rabbitescape.engine.Block;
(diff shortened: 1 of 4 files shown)


tttppp:
Ideally we should check the overhead in looking up the region all the time, and perhaps use a cache if it's expensive.

rabbit-escape-render/src/rabbitescape/render/WaterRegionRenderer.java
```diff
@@ -43,21 +29,23 @@
     private final WaterAnimation waterAnimation;
 
     public WaterRegionRenderer(
-        WaterRegion region, 
+        int x,
+        int y,
```
tttppp: This is the bulk of the change in the PR. Replace a copy of the WaterRegion (which sometimes gets outdated and so the graphics go weird) with coordinates so the WaterRegion can be looked up.

andybalaam:
I will fall off my chair if looking up the water region causes a performance problem.  Nice fix!  Fixes #605

colonelfazackerley:
Very nice. a much better idea than how I did it originally. Amazing how I cant see the simpler way until it's shown to me. Thanks @tttppp 

I have to object to the title. The WaterRegionRenderers still hold state (waterHeight, particles). Though this is only state for making things look pretty.

I also now understand the importance of separating formatting and substantive code changes into different diffs @andybalaam

PR #627, opened 2020-06-08
Editing Erros

carlosfinn:
1. class seperation, Factory pattern
2. Token.java, Token.java (LineProcessor.java)
3. To enhance the ease of adding the kind of token


andybalaam:
@carlosfinn thank you for contributing to Rabbit Escape!  These changes look like they might be useful, but you should definitely discuss refactorings with us before working on them - there may be good reasons for why the code is the way it is now.

Also, this change is too big for me to be able to review - it would be much easier if broken into smaller bits.

Please could you open issues for the things you think we should change so we can discuss them?  Try to make them as small as possible.  Then I'd like one pull request per issue, so we can merge the easy ones quickly and spend time getting the harder stuff right.

Thanks!

andybalaam:
No response from the author, so closing this.

